PR #3, opened 2020-09-21
Added PublishTrimmed to Impostor.Server

GerardSmit:
Added `<PublishTrimmed>true</PublishTrimmed>` to trim the release with IL Linker. More information can be found [here](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained).

**Size before trimming:** 66.4 MB
**Size after trimming:** 32.3 MB

I've tested the following: creating lobby, joining game, chatting, mini-games, finishing game and returning to lobby. Everything seems to be working fine and no required dependencies are being trimmed.

**Edit:** I've changed the Dockerfile as well so it supports trimming.

**Docker image size before:** 106MB
**Docker image size after:** 53.4MB


AeonLucid:
Thanks!! 👍

PR #7, opened 2020-09-24
:sparkles: added list of recent used ips in the client

### LINKED ISSUE #6: [Client] Save recent used server ips
In case I'm connecting to multiple servers now and then it'd be nice to have them saved so we can just select them from a list.

joaosombrio:
I've replaced the text input control for a combobox control and now everytime the user connects to a server, the ip used is saved in a simple .cfg file (ips separated by newline).
A button to clear the combo and the cfg file was also included.

resolves #6


src/Impostor.Client/Forms/FrmMain.cs
```diff
@@ -1,80 +1,105 @@
 ﻿using System;
 using System.Diagnostics;
-using System.Threading.Tasks;
+using System.IO;
+using System.Linq;
 using System.Windows.Forms;
 using Impostor.Client.Core;
 using Impostor.Client.Core.Events;
+using ErrorEventArgs = Impostor.Client.Core.Events.ErrorEventArgs;
```
joaosombrio, line 8: This is because there is an ErrorEventArgs in System.IO namespace already

PR #12, opened 2020-09-25
Create GameKeywords.cs

MrKrakenied:


AeonLucid:
Thanks but this code is already existing in the codebase. See https://github.com/AeonLucid/Impostor/blob/master/src/Impostor.Shared/Innersloth/Data/GameKeywords.cs.

Also please target the dev branch next time. Thats on me for not having contribution notes somewhere.

Issue #25, opened 2020-09-27
Server for ARM processors [Rpi4]


ProManiFest2122:
Just like in the title. Can you make version of server files (Docker) for ARM proccesors ?

marie6324:
This could theoretically be done.
@AeonLucid, any issues with publishing for `linux-arm64`?

cedricfarinazzo:
```
Unable to find image 'aeonlucid/impostor:latest' locally
latest: Pulling from aeonlucid/impostor
df20fa9351a1: Pull complete 
3e9798fe6e73: Pull complete 
123889c8c1d3: Pull complete 
b55e47f864c7: Pull complete 
6308710af54c: Pull complete 
17012af03a4e: Pull complete 
Digest: sha256:f2edc97abe93520b09cfcf727922cc9de5418a17ac428f94ecbff5b5743edb07
Status: Downloaded newer image for aeonlucid/impostor:latest
standard_init_linux.go:211: exec user process caused "exec format error"
```

`mcr.microsoft.com/dotnet/core/sdk:3.1` and `FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-alpine` are not compatible with ARM

PR #26, opened 2020-09-27
Dockerfile arm

cedricfarinazzo:
I added a dockerfile for arm proccesors.

#25

### DIFF
--- a/Dockerfile.arm
+++ b/Dockerfile.arm
@@ -6,17 +6,17 @@ COPY src/Impostor.Server/Impostor.Server.csproj ./src/Impostor.Server/Impostor.S
 COPY src/Impostor.Shared/Impostor.Shared.csproj ./src/Impostor.Shared/Impostor.Shared.csproj
 COPY submodules/Hazel-Networking/Hazel/Hazel.csproj ./submodules/Hazel-Networking/Hazel/Hazel.csproj
 
-RUN dotnet restore -r linux-musl-x64 ./src/Impostor.Server/Impostor.Server.csproj && \
-    dotnet restore -r linux-musl-x64 ./src/Impostor.Shared/Impostor.Shared.csproj && \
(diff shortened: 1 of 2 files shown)


GerardSmit:
Please target the `dev` branch. Also it seems you've used the Dockerfile from the `master` branch, which is outdated.
Check https://github.com/AeonLucid/Impostor/blob/dev/Dockerfile for the new Dockerfile.

cedricfarinazzo:
I have update the dockerfile. It should be good now.
Do I have to update github actions config?

Dockerfile.arm
```diff
@@ -0,0 +1,23 @@
...
+    dotnet restore -r linux-arm ./src/Impostor.Shared/Impostor.Shared.csproj && \
+    dotnet restore -r linux-arm ./submodules/Hazel-Networking/Hazel/Hazel.csproj
+
+# Copy everything else.
+COPY submodules/. ./submodules/
+COPY src/. ./src/
+RUN dotnet publish -c release -o /app -r linux-arm --no-restore ./src/Impostor.Server/Impostor.Server.csproj
+
+# Final image.
+FROM mcr.microsoft.com/dotnet/core/runtime-deps:latest
```
GerardSmit, line 19:
```suggestion
FROM mcr.microsoft.com/dotnet/core/runtime-deps:3.1-alpine
```

We don't really want to use `latest` here incase a breaking change gets be released in `runtime-deps`.

GerardSmit:
There is also a merging conflict because the source branche is from `master`.
If you pull the `dev` branch to your branch, you can merge it.

Dockerfile.arm
cedricfarinazzo: We cannot use the alpine version. I don't know if linux-arm runtime identifier works with alpine image. Maybe using 3.1 is a better solution.
```suggestion
FROM mcr.microsoft.com/dotnet/core/runtime-deps:3.1
```

PR #34, opened 2020-10-01
Add iOS instructions

fullonrager:


AeonLucid:
Is that correct? For someone else it was somewhere in `/var/mobile/Containers/Data/Application/B00BA1E2-8B59-4ABE-B06D-F939B85D008A/Documents`.

fullonrager:
iOS application bundle and data directories are labelled with unique IDs on each device, there is no one universal identifier for Among Us or any other App Store app. In `/var/mobile/Containers/Data/Application/`, Filza File Manager automatically labels each folder with their app name, hence why it is stated to use Filza.

PR #37, opened 2020-10-02
Add support for RPC events from message type 6

BrightSkyz:
This adds in support for getting RPC events from the message type 6 (gamedata)


src/Impostor.Server/Net/Messages/Rpc/RpcMessage13SendChat.cs
```diff
@@ -0,0 +1,15 @@
...
+using Impostor.Shared.Innersloth.Data;
+
+namespace Impostor.Server.Net.Messages.Rpc
+{
+    internal static class RpcMessage13SendChat
+    {
+
+        public static void Deserialize(MessageReader reader, out int rpcChatLength, out byte[] rpcChatMessage)
+        {
+            rpcChatLength = reader.ReadPackedInt32();
```
Mm2PL, line 11: Pretty sure this could be a `ReadString()`, could it? `ReadString()` also does decodes the message as Unicode, but that shouldn't be a problem.
```csharp
namespace Hazel
{
    public class MessageReader : IRecyclable
    {
    // [snip]
        public string ReadString()
        {
            int len = this.ReadPackedInt32();
            string output = UTF8Encoding.UTF8.GetString(this.Buffer, this.readHead, len);

            this.Position += len;
            return output;
        }
    // [snip]
    }
}
```

src/Impostor.Server/Net/Messages/Rpc/RpcMessageFlags.cs
```diff
@@ -0,0 +1,37 @@
...
+        public const byte SendChat = 13;
+        public const byte StartMeeting = 14;
+        public const byte SetScanner = 15;
+        public const byte SendChatNote = 16;
+        public const byte SetPet = 17;
+        public const byte SetStartCounter = 18;
+        public const byte EnterVent = 19;
+        public const byte ExitVent = 20;
+        public const byte SnapTo = 21;
+        public const byte Close = 22;
```
Mm2PL, line 27: q: What does `Close` mean here? Also are the names extracted from the game itself?

Mm2PL (approved):
I like these changes, however I've got questions. :)

AeonLucid:
Similar work is already in dev at the moment.

PR #38, opened 2020-10-02
Added Costura.Fody and changed build actions to make build a single .exe

LelouBil:
Impostor.Client : 
- Added Fody and Costura.Fody to package references into the exe.
 - Set App.config  build action to "Embedded Resource" to embed it into the exe.


AeonLucid:
Won't be using Fody, sorry.

PR #40, opened 2020-10-04
Add proton path detection

### LINKED ISSUE #11: Client nonfunctional with Among Us installed on Linux through Steam
The client does not function when attempting to patch a Steam installation of Among Us on Linux, complaining that the Among Us install directory doesn't exist.

I've tried executing the client with both Proton and straight Wine (from within Steam's wineprefix); it otherwise runs fine.

I added a symlink `drive_c/users/<username>` pointing to `drive_c/users/steamuser`, suspecting that it was looking at the wrong user directory and thus not finding AppData; this did absolutely nothing, surprisingly.

I have not tested in vanilla Wine completely without Steam as I do not own a DRM-free (itch.io) copy of Among Us.

This could easily be solved by simply adding a manual override for `regionInfo.dat`'s location; this would even allow the client to be run natively with mono on Linux instead of running within Wine.

js6pak:
Fixes #11
Migrates client's .csproj to new format allowing building it under mono.


AeonLucid:
Sorry but I don't want to merge csproj format until .NET 5 is released. 
GitHub CI has issues with .NET 5 in the new csproj format because the msbuild is outdated on the CI server.

js6pak:
Well, I can't build old csproj format soo? This PR can wait I guess.

src/Impostor.Client/Impostor.Client.csproj
```diff
@@ -1,102 +1,28 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
joaosombrio, line 1: I guess having the new csproj is great but why change the target framework though?

js6pak: `System.Resources.Extensions` (https://github.com/dotnet/msbuild/issues/4704) requires netstandard2.0 (net461)

AeonLucid:
Merged with #50.

PR #45, opened 2020-10-07
Reload config files when they change

### LINKED ISSUE #44: Config File must be reloaded at every system boot to set server IP
# Bug Report

## Base Information
- Raspberry Pi OS (ARM x86) - Headless
- Impostor 1.1.0
- Among Us v2020.9.22s

## I confirm:
- [x ] that I have searched for an existing bug report for this issue.

## Symptoms 

After rebooting the OS the server external IP address defaults to 127.0.0.1. Opening the config.json in nano and closing it resolves the issue, but it must be done after any reboot. 

## Reproduction

Reboot the Pi and start the server.

TheRealHona:
Closes #44

### DIFF
--- a/src/Impostor.Server/Program.cs
+++ b/src/Impostor.Server/Program.cs
@@ -54,8 +54,8 @@ namespace Impostor.Server
 #endif
                 .ConfigureAppConfiguration(builder =>
                 {
-                    builder.AddJsonFile("config.json", true);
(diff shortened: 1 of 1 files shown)


AeonLucid:
This is not the fix for that issue. 
Would also need to use `IOptionsMonitor` instead of `IOptions` for this PR to be useful.

PR #48, opened 2020-10-10
Update iOS instructions

fullonrager:
Added reference to the ImpostorConfig jailbreak tweak by pixelomer.


marie6324:
Have we verified that the tweak works?

fullonrager:
I tested it, it works.

PR #56, opened 2020-10-16
Fix AlterGame message flag

hplewis:
This should fix https://github.com/AeonLucid/Impostor/issues/55


hplewis:
screenshot of the client seeing the lobby is public

Screenshot taken by client immediately after entering lobby.

AeonLucid:
Thanks.. let's pretend this never happened okay?

PR #63, opened 2020-10-20
correct link

tuxinal:
the link to the web region file maker is returning a 404 error this link will do for now

PR #66, opened 2020-10-22
Update doc links

JorgenPhi:
Pulled in master to fix the issue tuxinal found (#63). Additionally, there were internal links that were not correct.


AeonLucid:
Thanks

PR #73, opened 2020-10-24
Fixed invalid path

JKamsker:
``Assembly.GetEntryAssembly()?.Location`` gives the path + Name.dll
``Path.Combine(rootFolder, "plugins")`` would lead to ``root\Assembly.dll\Plugins`` but we actually want ``root\plugins\``


AeonLucid:
Seems like this is indeed wrong, but this fix is not correct either.

> warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'

AeonLucid:
Fixed in https://github.com/Impostor/Impostor/commit/150dbdefaac3e2d23bdf2287ebc83dca15122961.

PR #90, opened 2020-10-27
Reload plugins on runtime

GerardSmit:
### Description
Added the functionality to reload plugins on runtime.

All the plugins run under a separate host which will be reloaded when the method`IPluginManager.ReloadAsync` is called. It's also possible to reload the assemblies with `IPluginManager.ReloadAsync(true)`.

In my testing reloading the `Impostor.Plugins.Example` works flawless. However when you reload the plugin `Impostor.Plugins.Debugger` (Blazor test plugin), the web server doesn't start after reloading the assemblies. I haven't had the time to look why.

#### Proxies
Because the plugins run under a different host, stopping and disposing the host will cause all the services to dispose. Because we register the API's from the server, we don't want them to be disposed. So we make sure that a proxy is being created that doesn't implement `IDisposable` (see the class `ProxyGenerator`).

#### Event Listeners
Currently the event listeners are not being detached when the plugins are being reloaded. I'll have to remove all the event listeners that were registered through the proxy, which I hopefully will do soon 😅.


AlexejheroYTB:
This PR will conflict with #145.

PR #93, opened 2020-10-27
Add PlayerExileEvent and PlayerMurderEvent

Saghetti0:
Adds PlayerExileEvent and PlayerMurderEvent. Note that PlayerExileEvent isn't called for now, because the Exiled RPC call never gets called.

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -273,9 +273,9 @@ namespace Impostor.Server.Net.Inner.Objects
                     if (!player.PlayerInfo.IsDead)
                     {
                         player.Die(DeathReason.Kill);
+                        await _eventManager.CallAsync(new PlayerMurderEvent(_game, sender, this, player));
                     }
 
-                    await _eventManager.CallAsync(new PlayerMurderEvent(_game, sender, this, player));
                     break;
                 }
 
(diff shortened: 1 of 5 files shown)


src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -273,6 +275,7 @@ public override async ValueTask HandleRpc(ClientPlayer sender, ClientPlayer? tar
                         player.Die(DeathReason.Kill);
                     }
 
+                    await _eventManager.CallAsync(new PlayerMurderEvent(_game, sender, this, player));
```
AeonLucid, line 278: Please put this into the if statement above.

AeonLucid (changes requested):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
Saghetti0: 1ad1a64a08250393ba51fc4771b7d238bc64aba9

codyphobe:
`Exile`, while not actually sent in-game, could be done via the [exiled player ID](https://github.com/Impostor/Impostor/blob/19cfad8132ca4ed35261d6a9908b6f44b0fae6d8/src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs#L79) in the `VotingComplete` RPC call on the `MeetingHud`.

PR #97, opened 2020-10-28
Added PlayerCompletedTaskEvent

danbudworthmead:
Added a PlayerCompletedTaskEvent when a player completes a task.


src/Impostor.Api/Events/Game/Player/IPlayerCompletedTaskEvent.cs
```diff
@@ -0,0 +1,7 @@
+namespace Impostor.Api.Events.Player
+{
+    public interface IPlayerCompletedTaskEvent : IPlayerEvent
+    {
+        uint TaskID { get; }
```
AlexejheroYTB, line 5: Perhaps we could have an enum?

PR #99, opened 2020-10-29
Player Reported Body Crash Fix

danbudworthmead:
There's a nullref exception thrown if a player reports a body the same second that a victim is killed.

I've added a nullptr check and updated the comments for the interface accordingly.


AeonLucid:
Why does this happen? Does the playercontrol get despawned when a player is killed?

danbudworthmead:
Not sure if I'm being totally honest ha

PR #100, opened 2020-10-29
Server shutdown nullref exception fix.

danbudworthmead:
Null reference check when closing down a server before disconnecting clients.


AeonLucid:
I think it might be better to figure out what causes the null reference. Do you have any stacktraces or code that can reproduce this issue consistently?

danbudworthmead:
> I think it might be better to figure out what causes the null reference. Do you have any stacktraces or code that can reproduce this issue consistently?

It seemed to be due to plugins not disposing properly when the server shuts down.

src/Impostor.Server/Events/MultiDisposable.cs
```diff
@@ -19,8 +19,8 @@ public void Dispose()
         {
             foreach (var disposable in _disposables)
             {
-                disposable.Dispose();
+                disposable?.Dispose();
```
AlexejheroYTB, line 22: This line is throwing NullReferenceExceptions in other cases, unrelated to shutting down the server. In my opinion, this change introducing the null conditional operator should be merged regardless if this pr is merged or not.

AeonLucid:
Added the null check to `dev`, rest of this PR is identical to https://github.com/Impostor/Impostor/pull/108.

Issue #103, opened 2020-10-30
Allow plugins to send game state changes


danbudworthmead:
# Api Suggestion

## Suggestion
Plugins should be able to send a packet changing all of the clients game states. 

## Use case
Having a custom victory / loss scenario. I currently achieve this but setting all clients to Impostor to end the round. 

## Expected result
Plugins able to set the game state and the clients act accordingly.

AeonLucid:
This is kind of a global request. There are a lot of packets that modify game state.

danbudworthmead:
> This is kind of a global request. There are a lot of packets that modify game state.

I meant more like having a Game.SetGamrState(GameStates.Fimished) function

AeonLucid:
But how do you propose the finished screen to show? There's not really a packet for it so we need to use hacky methods.

danbudworthmead:
> But how do you propose the finished screen to show? There's not really a packet for it so we need to use hacky methods.

Does MessageFlags.EndGame not put all the clients into an end of game state?

AlexejheroYTB:
> Does MessageFlags.EndGame not put all the clients into an end of game state?

Nope, as I described in my comment on #108:
 
> This causes everyone but the host to stop receiving packets, and does not end the game as intended.

I do not think there is a way to achieve this without client-side mods, unless what you said actually works:

> I currently achieve this by setting all clients to Impostor to end the round.

AlexejheroYTB:
Update: Sending an EndGame message could potentially work.

PR #108, opened 2020-11-02
Forcibly end the game

danbudworthmead:
### Description
Sets the game state to Ended and sets the players state to waiting for host.

### Closes issues
https://github.com/Impostor/Impostor/issues/103
- #


AlexejheroYTB (changes requested):
This causes everyone but the host to stop receiving packets, and does not end the game as intended.

PR #110, opened 2020-11-02
Fixed IsCancelled on interface events

GerardSmit:
### Description
Fixed exception when the an interface type extends `IEventCancelable` and is being used in `EventManager.CallAsync<T>`.


AeonLucid:
😍

PR #117, opened 2020-11-06
Make anticheat configurable

PandaDriver156:
### Description
Allows users to disable anti-cheat through the config file


js6pak:
This will cancel all logic after cheat attempt - events are going to break etc

AeonLucid:
Appreciate the PR but as @js6pak said this will break a lot, definitely not the way to do it. I want to move all the cheat exceptions to a single method which;
1. reads the config file
2. calls the plugin api
3. throws an exception

PR #123, opened 2020-11-09
Ban kill cooldown bypass

js6pak:


reecebenson (approved):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
```diff
@@ -37,6 +38,13 @@ public InnerPlayerInfo(byte playerId)
 
         public List<InnerGameData.TaskInfo> Tasks { get; internal set; }
 
+        public DateTimeOffset LastMurder { get; set; }
+
+        public bool CanMurder(IGame game)
+        {
+            return DateTimeOffset.UtcNow.Subtract(LastMurder).TotalSeconds < game.Options.KillCooldown;
+        }
```
AlexejheroYTB, line 46: This is a public method which returns the exact opposite of what it should return. Flip this with `>=`.

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -282,6 +282,13 @@ public override async ValueTask HandleRpc(ClientPlayer sender, ClientPlayer? tar
                         throw new ImpostorCheatException($"Client sent {nameof(RpcCalls.MurderPlayer)} as crewmate");
                     }
 
+                    if (sender.Character.PlayerInfo.CanMurder(_game))
```
AlexejheroYTB, line 285: Flip this as well with a `!`.

AlexejheroYTB (changes requested):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
js6pak: Whoops

AlexejheroYTB (approved):

PR #124, opened 2020-11-09
Console input

js6pak:


rhonda:
If I understand it correctly that would enable to potentially write plugins that allow server admins to kick users from server side directly, and potentially create some ban interface to blacklist some users to connect (like it they misbehave)? If so that would be a really groundbreaking change that enhances the usability by big times. <3

PR #129, opened 2020-11-10
Add GameOverReason

### LINKED ISSUE #114: Add end reason to game end event
# Api Suggestion

## Suggestion
My suggestion is to add an EndReason property to the IGameEndedEvent.

## Use case
This would allow plugins to know who won the game. Could be used for stat tracking? I'm not sure. My personal use case is modifying a player's role just before the game ends depending on which side won.

## Expected result

Name | Value
-- | -- 
HumansByVote | 0
HumansByTask | 1
ImpostorByVote | 2
ImpostorByKill | 3
ImpostorBySabotage | 4
ImpostorDisconnect | 5
HumansDisconnect | 6

I think 5 and 6 are unused. When the win is caused by disconnect, the value is either 0 or 2.

AlexejheroYTB:
### Description

This pull request adds a GameOverReason to the IGameEndedEvent

Closes #114

### DIFF
--- a/src/Impostor.Api/Net/Messages/C2S/Message08EndGameC2S.cs
+++ b/src/Impostor.Api/Net/Messages/C2S/Message08EndGameC2S.cs
@@ -1,0 +1,18 @@
+﻿using Impostor.Api.Innersloth;
+
+namespace Impostor.Api.Net.Messages.C2S
+{
+    public class Message08EndGameC2S
+    {
+        public static void Serialize(IMessageWriter writer)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public static void Deserialize(IMessageReader reader, out GameOverReason gameOverReason, out bool showAd)
+        {
+            gameOverReason = (GameOverReason)reader.ReadByte();
+            showAd = reader.ReadBoolean(); // Perhaps we could disable ads on Impostor?
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Impostor.Api/Net/Messages/C2S/Message08EndGameC2S.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static void Serialize(IMessageWriter writer)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public static void Deserialize(IMessageReader reader, out GameOverReason gameOverReason, out bool showAd)
+        {
+            gameOverReason = (GameOverReason)reader.ReadByte();
+            showAd = reader.ReadBoolean(); // Perhaps we could disable ads on Impostor?
```
reecebenson, line 15: I think it's in Impostor's best interest to leave that as it is. Otherwise we'd be stripping out a revenue income stream from Innersloth, and not sure what type of legal implications that would have.

reecebenson (approved):

AeonLucid:
Please make a PR from another branch than `dev` next time.

src/Impostor.Api/Net/Messages/C2S/Message08EndGameC2S.cs
AeonLucid: Good one, I agree. @reecebenson

PR #143, opened 2020-11-12
Add MapTypes and update SystemTypes

AlexejheroYTB:
This PR has two parts.

---

### Update SystemTypes
I've updated SystemTypes since multiple types were missing from the polus map.

### Add MapTypes
I've added a Map property of type MapTypes to the GameOptionsData class. I marked the MapId class as obsolete and replaced it with the MapTypes enum. The reason behind this is that newer plugin makers might not know about the existence of the MapId class, but if the property has an enum value instead, it won't go unnoticed. Since plugin makers already compare GameOptionsData.MapId to MapId.[Map], I see no harm in making it an enum.

---

Please leave your thoughts about the second change in the comments below. If it is unfit for Impostor, I will remove it from the PR.


AeonLucid:
The problem with the map types is that when you list a game, the maps are sent as a bit field through the GameOptionsData.MapId. This PR will break that I believe or have you thought about this?

AlexejheroYTB:
Oh right, I forgot about that. 😬 

Well, there is already a MapFlags enum, so what we could do is replace all bit field map id `int` values with `MapFlags` values.

AeonLucid:
We could make the MapId `internal` and add a `MapFlag` property as `internal` too, both properties write and read from `MapId` and make the nececessary conversions. Then keep the MapTypes one public to be used by plugins.

AlexejheroYTB:
The only place where `MapFlags` is publicly available is `GameManagerExtensions.GetGameCount`

Since `MapId` is cast into `MapFlags` only twice in the source code (and in different ways), I don't think a MapFlags property would be necessary.

PR #145, opened 2020-11-12
Add pre/post events and event cancellation

### LINKED ISSUE #92: Add event cancelling to chat messages
# Api Suggestion

## Suggestion
As the ability to cancel events with the API

## Use case
For my particular use case, I want to be able to cancel message events starting with `/` 

## Expected result
Player Sends Message Starting With "/",
The Event Is Cancelled,
No player receives the message
(Message can still be sent with e.Game.SendToAsync())

### LINKED ISSUE #135: Pre and post events; Cancellation API
### Current event call order

- In the current version of Impostor, certain events, such as `StartGame`, are called before the packet is sent to the clients, while others, such as `GameData` are called after the packets have been sent.

### Why this is bad (IMO)

- If there is no consistency, people can't know whether an event handler is ran before or after a packet without looking in the source code.

---

### Proposal 1

- An implementation of Pre and Post events (similar to Harmony Prefix and Postfix patches).
- A Pre event would run before the packet would be forwarded, while a Post event would run after it.
- A Pre event could be differentiated from a Post event via a new attribute, or a new parameter in the `EventPriority` attribute.
- Default would be Post.

### Advantages

- Easier implementation of event cancellation API for Pre events.
- Noobies doing single-threaded madness would no longer timeout packets such as `StartGame`.

---

### Proposal 2

- Moving all event handling before packet forwarding and allowing cancellation there.

### Advantages

- Less complexity in event handling

### Disadvantages

- Exception handling? Noobies doing single-threaded madness? Timeouts?

AlexejheroYTB:
### Description

This pull request adds Pre/Post events and event cancellation as described in the _Proposal 1_ section of #135.

Cancellable events may be cancelled by any pre event listener by changing the `Cancel` property on it. Similar to harmony patches, even if an event is cancelled, all pre event listeners are still run, and they have the ability to un-cancel the event. If after all of the event listeners have been called the `Cancel` property is false, the packet won't be sent and post event listeners won't be called.

---

<details>
<summary>Details (click to expand)</summary>

### Specifications

- New enum: `EventPriority`. Possible values: `Pre`, `Post`, `All`.
  - Event listeners for `Pre` will be called before the packet has been sent to other clients. Certain events can be cancelled in here. If the event doesn't have a `Pre` call step, `Pre` listeners will run on `Post` with a warning in the console.
  - Event listeners for `Post` will be called after the packet has been send to other clients.
  - Event listeners for `All` will be called on both `Pre` and `Post`. This is _useless._ I don't know why anyone would ever do this. This is mainly used internally to indicate that an event doesn't have a `Pre` call step.
- `EventPriorityAttribute` constructor can now take in an `EventCallStep` parameter. If none is provided, the default call step is `Post`.

---

### Example Usage

```cs
[EventPriority(EventCallStep.Pre)]
public ValueTask(IPlayerChatEvent e)
{
    if (e.Message.StartsWith("/")) e.Cancel = true;
}
```

---

### Event Changes

- `IGameAlterEvent`: pre/post, can be cancelled
  - sender will receive reverse packet _(TODO: make sure this works)_
  - other clients won't receive the packet
- `IGameCreatedEvent`: pre/post, can be cancelled _(TODO: catch exception in Client.cs)_
  - sender will be stuck in a black screen for 10s as if servers aren't available _(TODO: kick them instead)_
  - other clients won't receive the packet
- `IGameDestroyedEvent`: post, cannot be cancelled _(TODO: make sure this is actually post)_
- `IGameEndedEvent`: post, cannot be cancelled _(TODO: make sure this is actually post)_
- `IGamePlayerJoinedEvent`: pre/post, can be cancelled
  - sender will think that the game was destroyed _(TODO: should we use a custom message for this?)_
  - other clients won't receive the packet
- `IGamePlayerLeftEvent`: post, can't be cancelled _(TODO: make sure this is actually post)_
- `IGameStartedEvent`: post, cannot be cancelled _(TODO: make sure this is actually post)_
- `IGameStartingEvent`: pre/post, can be cancelled
  - idk what happens to the sender _(TODO: find out)_
  - other clients won't receive the packet

_I will add more events soon. Note that the changed files have some code for other events not listed above. Those are still work in progress, as is this whole pull request._

---

### Internal Changes

- `<IEventManager>.CallAsync` now has one additional parameter of type `EventCallStep`.
  - Specifying `EventCallStep.Pre` will run all event handlers listening on the `Pre` call step.
  - Specifying `EventCallStep.Post` will run all event handlers listening on the `Post` call step.
  - Specifying `EventCallStep.All` indicates that the event only has a single call step, which is run after packet forwarding. This will run all `Post` event listeners as usual, and will run all `Pre` event listeners with a warning in the log.
- `<IEventManager>.CallAsync` now returns `ValueTask<bool>` instead of `ValueTask`.
  - When running a `Pre` event, the return value will signify if the packet handling should continue or not. (`true` for keep going, `false` to cancel)
  - When running a `Post` or `All` event in a perfect world, the return value will always be `true`. Since we do not live in a perfect world, this return value will be `false` if someone tries to cancel in a post event. As a result, it is recommended to ignore this value.
- Renamed `IEventCancelable` to `ICancellableEvent`
- There seemed to be some preparations in the code for cancelling packets. These have been removed.
- Introduced a bunch of compilation errors which I need to fix. 😃

</details>

---

closes #92
closes #135


Lucaskyy:
This PR looks very nice! I would for sure use this. I hope you continue this 👍

src/Impostor.Server/Events/EventManager.cs
```diff
@@ -88,8 +89,39 @@ public async ValueTask CallAsync<T>(T @event)
 
                 foreach (var (handler, eventListener) in handlers)
                 {
-                    await eventListener.InvokeAsync(handler, @event, _serviceProvider);
+                    if (callStep == EventCallStep.All && eventListener.CallStep == EventCallStep.Pre)
+                    {
+                        _logger.LogWarning(
+                            "Event listener {0} for event {1} hall a call step of Pre, but the event does not support it. Running as Post instead.",
```
AeonLucid, line 95: hall > had

PR #164, opened 2020-11-16
Fix [API Bug] Task indices being cast to TaskTypes #159

### LINKED ISSUE #159: [API Bug] Task indices being cast to TaskTypes
This issue was reported by `Cybershard#3935` on Discord.

---

> https://github.com/Impostor/Impostor/blob/d590da0d27388e5b95db9c7bc7ba3558b1a57135/src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs#L69-L76 Unless I'm missing something, this code gets called when RPC updategamedata or the gamedata object's deserialize method gets called, but in any case, **the reader contains the index of the task for that InnerPlayerInfo, and if its completed or not. Right now, its being treated as a TaskType**

> Same goes for this code: https://github.com/Impostor/Impostor/blob/d590da0d27388e5b95db9c7bc7ba3558b1a57135/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs#L83-L85

> The id from the RPC CompleteTask is the index of what task the player completed in regards to their InnerPlayerInfo. But, the code at https://github.com/Impostor/Impostor/blob/d590da0d27388e5b95db9c7bc7ba3558b1a57135/src/Impostor.Server/Events/Game/Player/PlayerCompletedTaskEvent.cs#L11-L17 is casting it as a TaskType.

### LINKED ISSUE #194: PlayerCompletedTaskEvent Task is wrong
/src/Impostor.Server/Events/Game/Player/PlayerCompletedTaskEvent.cs

Task reports back the wrong task completed.

First task completed seems to be submitScan which is 0, and the number seems to increment upward from there through the task list skipping a few along the way, but still incrementing. 

I stopped at 14 (Divert Power) 

My testing setup: 
Add logging to PlayerCompletedTaskEvent after Task = (TaskTypes)taskID; 
System.Console.WriteLine($"{playerControl.PlayerInfo.PlayerName} completed {Task.ToString()}");

Set common tasks to 30 (this gives a bunch of swipe card I can do back to back) 

It's more likely that taskID is not the task, but an index of that players task, which means Task = (TaskTypes)taskID; returns the wrong task. It could be more like Task = (TaskTypes)player.Character.Tasks[taskID] (would need to check, but you get the idea)

cybershard:
### Description
In the game, RPC SetTasks gets called with the tasks array where the taskId is in an enum, and then RPC UpdateGameData with the index of the task with respect to a `InnerPlayerInfo` and whether if its complete or not. The code now updates the `Complete` boolean by index, instead of creating a new `TaskInfo`
### Closes issues

- closes #159 
- closes #194

### DIFF
--- a/src/Impostor.Api/Net/Inner/Objects/ITaskInfo.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/ITaskInfo.cs
@@ -5,7 +5,7 @@ namespace Impostor.Api.Net.Inner.Objects
 {
     public interface ITaskInfo
     {
-        uint TaskIndex { get; }
+        uint Id { get; }
 
         TaskTypes Type { get; }
 
--- a/src/Impostor.Api/Net/Inner/Objects/ITaskInfo.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/ITaskInfo.cs
@@ -5,8 +5,10 @@ namespace Impostor.Api.Net.Inner.Objects
 {
     public interface ITaskInfo
     {
+        uint TaskIndex { get; }
+
         TaskTypes Type { get; }
-        
+
         bool Complete { get; }
     }
 }
--- a/src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
@@ -1,4 +1,5 @@
-﻿using Impostor.Api.Innersloth;
+﻿using System.Threading.Tasks;
+using Impostor.Api.Innersloth;
 using Impostor.Api.Net.Inner.Objects;
 using Impostor.Api.Net.Messages;
 
@@ -8,6 +9,8 @@ namespace Impostor.Server.Net.Inner.Objects
     {
         public class TaskInfo : ITaskInfo
         {
+            public uint TaskIndex { get; internal set; }
+
             public TaskTypes Type { get; internal set; }
 
             public bool Complete { get; internal set; }
@@ -20,7 +23,7 @@ namespace Impostor.Server.Net.Inner.Objects
 
             public void Deserialize(IMessageReader reader)
             {
-                reader.ReadPackedUInt32();
+                TaskIndex = reader.ReadPackedUInt32();
                 this.Complete = reader.ReadBoolean();
             }
         }
--- a/src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
@@ -20,7 +20,7 @@ namespace Impostor.Server.Net.Inner.Objects
 
             public void Deserialize(IMessageReader reader)
             {
-                this.Type = (TaskTypes)reader.ReadPackedUInt32();
+                reader.ReadPackedUInt32();
                 this.Complete = reader.ReadBoolean();
             }
         }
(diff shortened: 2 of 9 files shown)


src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
```diff
@@ -20,9 +20,9 @@ public void Serialize(IMessageWriter writer)
 
             public void Deserialize(IMessageReader reader)
             {
-                this.Type = (TaskTypes)reader.ReadPackedUInt32();
+                reader.ReadPackedUInt32();
```
AlexejheroYTB, line 23: Maybe the index should be saved as well in case the plugin maker wants to do something with it.

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -81,7 +81,10 @@ public override async ValueTask HandleRpc(ClientPlayer sender, ClientPlayer? tar
                     }
 
                     var taskId = reader.ReadPackedUInt32();
-                    await _eventManager.CallAsync(new PlayerCompletedTaskEvent(_game, sender, this, taskId));
+                    var task = PlayerInfo.Tasks[(int)taskId] ??= new InnerGameData.TaskInfo();
```
AlexejheroYTB, line 84: Maybe throw a warning if the task doesn't exist?

src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
cybershard: Index is already saved as the actual index in the List<TaskInfo>. Should we create another property for task index in TaskInfo?

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
cybershard: I will throw a warning if the player tries to complete a task they don't have or complete a task they have already done.

src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
AlexejheroYTB: Yes, so it can be used in events where only an ITaskInfo is provided.

src/Impostor.Api/Net/Inner/Objects/ITaskInfo.cs
```diff
@@ -5,8 +5,10 @@ namespace Impostor.Api.Net.Inner.Objects
 {
     public interface ITaskInfo
     {
+        public uint TaskIndex { get; }
```
AlexejheroYTB, line 8: Access modifiers in interfaces? 🤔

cybershard: Thanks

src/Impostor.Server/Net/Inner/Objects/InnerGameData.TaskInfo.cs
```diff
@@ -20,9 +23,9 @@ public void Serialize(IMessageWriter writer)
 
             public void Deserialize(IMessageReader reader)
             {
-                this.Type = (TaskTypes)reader.ReadPackedUInt32();
+                TaskIndex = reader.ReadPackedUInt32();
                 this.Complete = reader.ReadBoolean();
             }
```
AlexejheroYTB, line 28: The deserialize method doesn't assign a tasktype, so it might be missing in some cases. Maybe get the tasktype from the player info in here somehow?

cybershard: No, because it is set for the player info by RPC set tasks, where it sets it here as well (RPC set tasks and RPC game data is sent together). For the API, a SetTasks API call should update the player info tasks and the task indexes itself.

reecebenson (approved):

AlexejheroYTB (approved):
Edit the description of this PR and add `closes #194`

PR #170, opened 2020-11-19
expose player position to api

### LINKED ISSUE #167: Is there a way to get the player current position on the map?
I was trying to do a plugin that changes a player position with other player
I was able to change the player position with player.Character.NetworkTransform.SnapToAsync(newPosition), but I wanted to know if it is possible to know all the players position at a given time, something like this:

public void OnGameStarted(IGameStartedEvent e)
{
  _logger.LogInformation($"Game is starting.");

  foreach (var player in e.Game.Players)
    {
       _logger.LogInformation($"{player.Character.NetworkTransform.CurrentPlayerPosition}") // here we know the player position
     }
}

Also, I did not see how people get the Map coordinates limits (like, in Skeld which coordinates the Admin/Cafeteria is, something like that). Does anyone know that?

Also, is there a way to allow the chat use during the game phase (not only in the discussion phase)? So the users can type a message like:
/teleport (2.5, -1.53)
and be teleported to that location (using the message listener).

Anyway, thanks for the attention. Great Api!

propstg:
### Description
Expose _targetSyncPosition to API, so plugins can get player position.

### Closes issues

- closes #167 (partially?)

### DIFF
--- a/src/Impostor.Api/Net/Inner/Objects/Components/IInnerCustomNetworkTransform.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/Components/IInnerCustomNetworkTransform.cs
@@ -11,5 +11,11 @@ namespace Impostor.Api.Net.Inner.Objects.Components
         /// <param name="position">The target position.</param>
         /// <returns>Task that must be awaited.</returns>
         ValueTask SnapToAsync(Vector2 position);
+
+        /// <summary>
+        ///     Gets the current position <see cref="IInnerPlayerControl"/>.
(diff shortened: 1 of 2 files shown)


AeonLucid:
This is wrong and precisely the reason why position has not been added to the api yet. It is a target position, a prediction, where it is going to be at a later time.

I'm still unsure how to implement this properly.

propstg:
Oh, interesting. I guess that makes sense. Thanks for the explanation!

PR #171, opened 2020-11-19
Add PlayerMovmentEvent Pooling and Position at NetworkTransform

### LINKED ISSUE #167: Is there a way to get the player current position on the map?

ZiadMontaser:
### Description
I Added IPLayerMovmentEvent at Impostor.Api;
and Imlemented pooling for PLayerMovmentEvent;
and Added Position Property at IInnerCustomNetworkTransform to access the position from the plugins
<!-- 

### Closes issues
closes #167
- closes #

### DIFF
--- a/src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
+++ b/src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Threading.Tasks;
 using Impostor.Api.Events;
 
 namespace Impostor.Plugins.Example.Handlers
@@ -24,7 +25,7 @@ namespace Impostor.Plugins.Example.Handlers
         }
 
         [EventListener]
-        public void OnGameStarted(IGameStartedEvent e)
+        public async void OnGameStarted(IGameStartedEvent e)
         {
             Console.WriteLine("Game > started");
 
@@ -33,6 +34,12 @@ namespace Impostor.Plugins.Example.Handlers
                 var info = player.Character.PlayerInfo;
 
                 Console.WriteLine($"- {info.PlayerName} {info.IsImpostor}");
+            }
+            await Task.Delay(15 * 1000);
+            foreach(var player in e.Game.Players)
+            {
+                Console.WriteLine($"{player.Character.PlayerInfo.PlayerName} moved to {player.Character.NetworkTransform.Position}");
+
             }
         }
 
--- a/src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
+++ b/src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
@@ -35,12 +35,6 @@ namespace Impostor.Plugins.Example.Handlers
 
                 Console.WriteLine($"- {info.PlayerName} {info.IsImpostor}");
             }
-            await Task.Delay(15 * 1000);
-            foreach(var player in e.Game.Players)
-            {
-                Console.WriteLine($"{player.Character.PlayerInfo.PlayerName} moved to {player.Character.NetworkTransform.Position}");
-
-            }
         }
 
         [EventListener]
--- a/src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
+++ b/src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
@@ -1,13 +1,21 @@
-﻿using Impostor.Api.Events.Player;
+using System;
+using System.Collections.Concurrent;
+
+using Impostor.Api.Events.Player;
 using Impostor.Api.Games;
 using Impostor.Api.Net;
 using Impostor.Api.Net.Inner.Objects;
+using Impostor.Hazel;
 
 namespace Impostor.Server.Events.Player
 {
     // TODO: Finish and use event, needs to be pooled
-    public class PlayerMovementEvent : IPlayerEvent
+    public class PlayerMovementEvent : IPlayerMovementEvent , IRecyclable , IDisposable
     {
+        private static readonly ConcurrentBag<PlayerMovementEvent> items = new ConcurrentBag<PlayerMovementEvent>();
+        private static int counter = 0;
+        private static int MAX = 25;
+
         public PlayerMovementEvent(IGame game, IClientPlayer clientPlayer, IInnerPlayerControl playerControl)
         {
             Game = game;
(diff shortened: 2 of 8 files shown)


src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
```diff
@@ -1,24 +1,71 @@
...
-        public IGame Game { get; }
+        public IGame Game { get; private set; }
+
+        public IClientPlayer ClientPlayer { get; private set; }
 
-        public IClientPlayer ClientPlayer { get; }
+        public IInnerPlayerControl PlayerControl { get; private set; }
 
-        public IInnerPlayerControl PlayerControl { get; }
+        public static PlayerMovementEvent Get(IGame game, IClientPlayer clientPlayer, IInnerPlayerControl playerControl)
```
AlexejheroYTB, line 32: Could there be a better method name for this? Also, why is this a static method?

src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.Api.cs
```diff
@@ -6,6 +6,8 @@ namespace Impostor.Server.Net.Inner.Objects.Components
 {
     internal partial class InnerCustomNetworkTransform : IInnerCustomNetworkTransform
     {
+        public Vector2 Position => _targetSyncPosition;
```
AlexejheroYTB, line 9: Is this correct? Isn't this still pointing to the target position?

ZiadMontaser: Yes, isn't _targetSyncPosition contains the player position?

src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
ZiadMontaser: This method checks if there is an existing instance of PLayerMovmentEvent to use it instead of constructing a new object.
It is static to be easier to use anywhere.

src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.Api.cs
AlexejheroYTB: AFAIK _targetSyncPosition is the target position, not the current one.

src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
```diff
@@ -24,7 +25,7 @@ public void OnGameStarting(IGameStartingEvent e)
         }
 
         [EventListener]
-        public void OnGameStarted(IGameStartedEvent e)
+        public async void OnGameStarted(IGameStartedEvent e)
```
reecebenson, line 28: Why is this `async` when its not using any asynchronous tasks?

src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
```diff
@@ -1,24 +1,71 @@
...
 using Impostor.Api.Games;
 using Impostor.Api.Net;
 using Impostor.Api.Net.Inner.Objects;
+using Impostor.Hazel;
 
 namespace Impostor.Server.Events.Player
 {
     // TODO: Finish and use event, needs to be pooled
-    public class PlayerMovementEvent : IPlayerEvent
+    public class PlayerMovementEvent : IPlayerMovementEvent , IRecyclable , IDisposable
```
reecebenson, line 13: Refactor: `IPlayerMovementEvent, IRecyclable, IDisposable` (remove unnecessary spaces)

src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
```diff
@@ -1,24 +1,71 @@
...
+            {
+                items.Add(this);
+                counter++;
+            }
+        }
+
+        public void Dispose()
+        {
+            Recycle();
+        }
```
reecebenson, line 69: `public void Dispose() => Recycle();`

src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
```diff
@@ -1,24 +1,71 @@
...
 
 namespace Impostor.Server.Events.Player
 {
     // TODO: Finish and use event, needs to be pooled
-    public class PlayerMovementEvent : IPlayerEvent
+    public class PlayerMovementEvent : IPlayerMovementEvent , IRecyclable , IDisposable
     {
+        private static readonly ConcurrentBag<PlayerMovementEvent> items = new ConcurrentBag<PlayerMovementEvent>();
+        private static int counter = 0;
+        private static int MAX = 25;
```
reecebenson, line 17: A more identifiable name, i.e. `MAX_MOVEMENT_STEPS` or something? (Assuming this is used to hold _X_ amount of previous movements, which could prove quite useful for anticheats.). This could also be a `const` instead of `static`.

reecebenson (changes requested):

src/Impostor.Server/Events/Game/Player/PlayerMovementEvent.cs
```diff
@@ -1,24 +1,71 @@
...
+using Impostor.Hazel;
 
 namespace Impostor.Server.Events.Player
 {
     // TODO: Finish and use event, needs to be pooled
-    public class PlayerMovementEvent : IPlayerEvent
+    public class PlayerMovementEvent : IPlayerMovementEvent , IRecyclable , IDisposable
     {
+        private static readonly ConcurrentBag<PlayerMovementEvent> items = new ConcurrentBag<PlayerMovementEvent>();
+        private static int counter = 0;
```
reecebenson, line 16: Should `counter` be static here? Or local to the classes instance, as I'd assume `X` amount of movements per player would be much more useful, instead of `X` amount of movements in the game?

src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.Api.cs
reecebenson: Yes, it is the target position, not the current player position. #170's PR was closed for using the same stuff.

AeonLucid:
Not going to merge this. It is unnecessary to implement pooling yourself because we can make use of `Microsoft.Extensions.ObjectPool`. Also see this comment https://github.com/Impostor/Impostor/pull/170#issuecomment-730188426

src/Impostor.Plugins.Example/Handlers/GameEventListener.cs
ZiadMontaser: I was using it for testing and forgot to remove it

src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.Api.cs
ZiadMontaser: can I ask a question what is the target position?

PR #180, opened 2020-11-22
[Docs] Add docker-compose instructions

hmmbob:
### Description
Adds instructions on how to run this server using docker-compose.
Removed unused badge for "edge" branch

### DIFF
--- a/docs/Running-the-server.md
+++ b/docs/Running-the-server.md
@@ -18,7 +18,6 @@ There are currently two modes to run the Impostor server in. The first way is th
 ### Using docker
 
 [![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor?sort=semver)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
-[![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor/edge)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
 
 ```
 docker run -p 22023:22023/udp aeonlucid/impostor:nightly
--- a/docs/Running-the-server.md
+++ b/docs/Running-the-server.md
@@ -18,6 +18,7 @@ There are currently two modes to run the Impostor server in. The first way is th
 ### Using docker
 
 [![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor?sort=semver)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
+[![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor/nightly)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
 
 ```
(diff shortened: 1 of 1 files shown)


dimaguy:
I was wondering if you could leave a note too on how do put in plugins in a shared folder with your method. Because otherwise it would be just another way to do it and thats too similar with the docker oneliner

hmmbob:
Sure, but I don't use that option. Got some docs on how this works? Is it just mounting the folder `plugins` inside the `/app` folder? Because that's easy: one can mount any dir into the container with such compose:

```
version: '3.4'
services:
  impostor:
    image: aeonlucid/impostor:nightly
    container_name: impostor
    ports:
      - 22023:22023/udp
    volumes:
      - /path/to/plugins/on/local/pc:/app/plugins
```

dimaguy:
Ye thats basically it. it just needs to be the ``root folder/plugins`` where impostor is on :)

hmmbob:
Done

docs/Running-the-server.md
```diff
@@ -18,12 +18,26 @@ There are currently two modes to run the Impostor server in. The first way is th
...
+
+services:
+  impostor:
+    image: aeonlucid/impostor:nightly
+    container_name: impostor
+    ports:
+      - 22023:22023/udp
+    volumes:
+      - /path/to/local/config.json:/app/config.json # For easy editing of the config
+      - /path/to/local/plugins:/app/plugins # Only needed if using plugins
```
reecebenson, line 38: Could we also add libraries, if that's the case?
`/path/to/local/libraries:/app/libraries  # Only needed if using external libraries`

hmmbob: Sure!

reecebenson (approved):

dimaguy (approved):

docs/Running-the-server.md
```diff
@@ -18,12 +18,27 @@ There are currently two modes to run the Impostor server in. The first way is th
 ### Using docker
 
 [![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor?sort=semver)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
-[![Docker Image](https://img.shields.io/docker/v/aeonlucid/impostor/edge)](https://hub.docker.com/repository/docker/aeonlucid/impostor)
```
AeonLucid, line 21: Why has this been removed? Please update to `https://img.shields.io/docker/v/aeonlucid/impostor/nightly`.

AeonLucid (changes requested):

docs/Running-the-server.md
hmmbob: It errored at 'edge' and the nightly was shown in the semver variant, hence why i removed it. It's back 👍

PR #182, opened 2020-11-23
Adding --errors-only argument

danbudworthmead:
Along with --verbose, it might be useful to have a --errors-only arg.

We use it on skeld.net so I thought I'd PR it as well.

### DIFF
--- a/src/Impostor.Server/Net/Manager/ClientManager.cs
+++ b/src/Impostor.Server/Net/Manager/ClientManager.cs
@@ -19,7 +19,8 @@ namespace Impostor.Server.Net.Manager
         private static HashSet<int> SupportedVersions { get; } = new HashSet<int>
         {
             GameVersion.GetVersion(2020, 09, 07), // 2020.09.07 - 2020.09.22
-            GameVersion.GetVersion(2020, 10, 08), // 2020.10.08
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Program.cs
```diff
@@ -40,6 +40,10 @@ private static int Main(string[] args)
             {
                 logLevel = LogEventLevel.Verbose;
             }
+            else if (args.Contains("--errors-only"))
+			{
```
reecebenson, line 44: Can you get rid of the tabs here, and replace with spaces?

src/Impostor.Server/Net/Manager/ClientManager.cs
```diff
@@ -19,7 +19,8 @@ internal partial class ClientManager
         private static HashSet<int> SupportedVersions { get; } = new HashSet<int>
         {
             GameVersion.GetVersion(2020, 09, 07), // 2020.09.07 - 2020.09.22
-            GameVersion.GetVersion(2020, 10, 08), // 2020.10.08
+            GameVersion.GetVersion(2020, 10, 08), // 2020.10.08            
+            GameVersion.GetVersion(2020, 11, 17), // 2020.11.17
```
reecebenson, line 23: Can you rebase?

reecebenson (changes requested):

Issue #183, opened 2020-11-23
SetMurderedAsync not working when called from player.Character


NickNish09:
I was trying to make a plugin where when you send a message with /kill player_name it would kill the player.
For some reason I was only able to use the .SetMurderedAsync when calling it from PlayerControl, from the event of a player chat:

```
...
public void OnPlayerChat(IPlayerChatEvent e)
        {
            string[] command = e.Message.Split(' ');
            if(command[0] == "/kill"){
                if(1 < command.Length){
                    IClientPlayer p = _room.GetPlayers().playerByName(command[1]); // returns the IClientPlayer with that PlayerName
                    if(p != null){
                        _logger.LogInformation($"{p.Character.PlayerInfo.PlayerName} will be murdered.");
                        e.PlayerControl.SetMurderedAsync(); // this works, but kills the actual player who sent the message, and only if he was an impostor
                        p.Character.SetMurderedAsync(); // this does't work
                        p.Character.SendChatAsync("I can send messages"); // did this only to check if p.Character.Something could still be executed. This works, the player sends the message.
                    }
                }
            }
        }
```

It seems that SetMurderedAsync() only works if called from e.PlayerControl (from IPlayerChatEvent), I was not able to use it with other players from the game. Is this the normal behavior? Or am I doing something wrong?
Thank you for your attention.

Mr-VIT:
`SetMurderedAsync` works only for Imposters as suicide

NickNish09:
I see, so there is no way to kill a player by chat commands right?
Let me ask one question if you mind, is there anyway to forcebily end the game and make the imposters win? I was trying to kill all the tripulants (with no success) and now I was trying to set everyone as infected (with no success either).

```
List<IInnerPlayerControl> toInfect = new List<IInnerPlayerControl>();
foreach(var p in e.Game.Players)
  {
     if(!p.Character.PlayerInfo.IsImpostor){
       toInfect.Add(p.Character);
     }
   }
e.Game.SetInfectedAsync(toInfect);
```
Tried to do this but the SetInfectedAsync doesnt seem to make anyone imposter.

js6pak:
1. You can't use `SetInfectedAsync` (it can be called only once by host, or it will break stuff), I have no idea why is it public.
2. To end the game you can just kill all impostors, or maybe using RpcEndGame if #103 gets resolved.

Please close the issue if you already resolved your problem.

PR #192, opened 2020-11-25
Fix set murdered

ZiadMontaser:
### Description
Fix set murdered
<!--

### DIFF
--- a/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
@@ -106,13 +106,6 @@ namespace Impostor.Api.Net.Inner.Objects
         ValueTask SendChatToPlayerAsync(string text, IInnerPlayerControl? player = null);
 
         /// <summary>
-        ///     Sets the current to be murdered <see cref="IInnerPlayerControl"/>.
-        ///     Visible to all players.
-        /// </summary>
-        /// <returns>Task that must be awaited.</returns>
-        ValueTask SetMurderedAsync();
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


js6pak:
Wouldn't this break game state, anticheat would think that player is still alive? And this really should be done other way around + with dead/impostor checks

ZiadMontaser:
are we good now

src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
```diff
@@ -111,5 +111,13 @@ public interface IInnerPlayerControl : IInnerNetObject
         /// </summary>
         /// <returns>Task that must be awaited.</returns>
         ValueTask SetMurderedAsync();
```
AlexejheroYTB, line 113: Remove `SetMurderedAsync`

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
```diff
@@ -103,5 +106,19 @@ public async ValueTask SetMurderedAsync()
             writer.Write((byte)NetId);
             await _game.FinishRpcAsync(writer);
         }
+
+        public async ValueTask SetMurderedByAsync(IClientPlayer impostor)
+        {
+            var impostorNetId = impostor.Character.NetId;
```
AlexejheroYTB, line 112: Check if the `impostor` is actually an impostor, check if they're not dead, and also check if the target player is not dead.

AlexejheroYTB (changes requested):

AlexejheroYTB:
Edit the description of the PR to include `closes #183`

PR #203, opened 2020-11-30
Save Hostname

### LINKED ISSUE #196: Patcher Client don't accept host names.
# Bug Report

## Base Information
- W10 Pro
- Impostor v1.2.2
- Among Us v2020.11.17s

## Reproduction

Open Impostor.exe
Put in Ip Adress some hostname that redirects to the ip: myamongus.tk
Click Save Button
It will save the latest ip of the list, not my hostname: myamongus.tk

Any way to fix this, please?

ty!!!

gavinsteinhoff:
### Description

Saves the hostname, if the user entered one. 

If the client enters a hostname, the program will pass it to other methods. When it comes to saving, the program will check if host is empty. If it is not, it will save the host instead of the IP. If is is empty, the program will save the IP.

### Closes issues

- closes #196


AeonLucid:
What about loading?

gavinsteinhoff:
> What about loading?

It is stored just like an IP, so the form will load it just the same and add the hostname to the dropdown list. Since everything is strings, it doesn't matter.

PR #211, opened 2020-12-05
[API] Added missing data to IGamePlayerJoinedEvent and IGamePlayerLeftEvent

### LINKED ISSUE #210: PlayerJoined and PlayerLeft are missing data
# Api missing data

## Data

[IGamePlayerLeftEvent ](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Api/Events/Game/IGamePlayerLeftEvent.cs) and [IGamePlayerJoinedEvent ](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs) do not provide methods to get the player
Even tho their implementations got such methods, see: 
[GamePlayerLeftEvent](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Server/Events/Game/GamePlayerLeftEvent.cs)
[GamePlayerJoinedEvent](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Server/Events/Game/GamePlayerJoinedEvent.cs)

## Use-case

To get the player that joined or left

Lauriichan:
### Description

Add the possibility to get the Player that left or joined

### Closes issues

- closes #210

### DIFF
--- a/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
+++ b/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
@@ -1,4 +1,6 @@
-﻿namespace Impostor.Api.Events
+using Impostor.Api.Net;
+
+namespace Impostor.Api.Events
 {
     public interface IGamePlayerJoinedEvent : IGameEvent
     {
--- a/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
+++ b/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
@@ -1,4 +1,5 @@
 using Impostor.Api.Net;
+using Impostor.Api.Net.Inner.Objects;
 
 namespace Impostor.Api.Events
 {
@@ -6,6 +7,7 @@ namespace Impostor.Api.Events
     {
         /// <summary>
         ///     Gets the <see cref="IClientPlayer"/> which triggered the event.
+        ///     The <see cref="IInnerPlayerControl"/> Character of this object is null since the Character object didn't spawn yet.
         /// </summary>
         IClientPlayer Player { get; }
     }
--- a/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
+++ b/src/Impostor.Api/Events/Game/IGamePlayerJoinedEvent.cs
@@ -2,5 +2,9 @@
 {
     public interface IGamePlayerJoinedEvent : IGameEvent
     {
+        /// <summary>
+        ///     Gets the <see cref="IClientPlayer"/> which triggered the event.
+        /// </summary>
+        IClientPlayer Player { get; }
     }
 }
--- a/src/Impostor.Api/Events/Game/IGamePlayerLeftEvent.cs
+++ b/src/Impostor.Api/Events/Game/IGamePlayerLeftEvent.cs
@@ -2,5 +2,9 @@
 {
     public interface IGamePlayerLeftEvent : IGameEvent
     {
+        /// <summary>
+        ///     Gets the <see cref="IClientPlayer"/> which triggered the event.
+        /// </summary>
+        IClientPlayer Player { get; }
     }
 }
--- a/src/Impostor.Api/Events/Game/IGamePlayerLeftEvent.cs
+++ b/src/Impostor.Api/Events/Game/IGamePlayerLeftEvent.cs
@@ -6,5 +6,10 @@
         ///     Gets the <see cref="IClientPlayer"/> which triggered the event.
         /// </summary>
         IClientPlayer Player { get; }
+        
+        /// <summary>
+        ///     Gets the the info if the player was banned or just left
+        /// </summary>
+        bool IsBan { get; }
     }
 }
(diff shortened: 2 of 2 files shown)


Mr-VIT (approved):

reecebenson (approved):

reecebenson:
Just a quick note, if somebody could confirm this... could we verify whether the `.Client` or `.Character` parameters of the `ClientPlayer` object may be null (if I remember correctly from when I did similar changes).

I believe `.Character` is unavailable for both (the player hasn't "Spawned" yet for `GamePlayerJoined` and the `.Character` is null when the player leaves as it has been disposed). This was weeks ago testing, so scenarios may have changed, but confirmation of this would be good.

LorenzoPapi:
`.Character` is null for `IGamePlayerJoinedEvent`, just tried with Impostor 1.2.2-ci.121. Though, I can't understand when the `IGamePlayerLeftEvent` fires, when I join a room and then leave the event doesn't fire.

realfaceitkantesilvestrs:
@Lauriichan I tested this out manually a week ago - it seems that data is just null most of the time - argo there is no benefit adding these to the list.

I just manually implemented local player tracker, and then just ran comparison of `.Game.Players` to figure out which players have left/joined.

Lauriichan:
> @Lauriichan I tested this out manually a week ago - it seems that data is just null most of the time - argo there is no benefit adding these to the list.
> 
> I just manually implemented local player tracker, and then just ran comparison of `.Game.Players` to figure out which players have left/joined.

Okay, yeah then I'll need to take that work around too I guess, I hope there will be a better solution then this later on ^^

AeonLucid:
Character is null on the join event and only set in the spawn event.

Lauriichan:
As long as I got access to for example the players IP Address from there and maybe it's name (isn't really needed) then I'm pretty fine with the event but maybe I should add a note to the documentation that the Character is null since yeah, hasn't spaned yet

Lauriichan:
Okay so if nobody has anything against those changes then I think this can be merged

kantesilvestrs:
@Lauriichan  I don't see the point of this change - what is the actual information you are getting do distinguish this player from any other player joining?

Lauriichan:
The players Ip

kantesilvestrs:
@Lauriichan players IP address is not unique especially in populated areas where IP is shared

Why `IPlayerSpawned` is not good enough?

Lauriichan:
Isn't the IPlayerSpawnedEvent called multiple times?

Lauriichan:
And also if you can't receive any information from those events then why are they there?
For what you would need those events if you can't even get any information of them?

kantesilvestrs:
@Lauriichan  you are getting `.Game` information

Lauriichan:
Yeah but why do you provide for example the Player and IsBan getter in the original event if they aren't available for the API event?

kantesilvestrs:
@Lauriichan imho it's just wishful thinking from the person who initially created that event and that is it - you can even see in the comments stating that the player information is not available

kantesilvestrs:
@Lauriichan believe you me -> I also want this event to work with Character information so that I can manage info before they spawn -> but you can't.

And I don't think a half solution is better than no solution in this situation

Lauriichan:
I think that it is enough information tho and pretty useful so yeah, but if you think that its not needed then just yeah don't merge it I guess but why remove for example the "IsBan" option from the IGamePlayerLeftEvent it's pretty useful so

Lauriichan:
And yeah, of course, you don't have the Character which would be pretty useful to have but even tho the information you got in the Connection of the player can also be pretty useful on join and leave, so in my opinion its pretty useful to have those informations

kantesilvestrs:
@Lauriichan Maybe instead of changing API interfaces you can look into to the actual events being generated and see if the Character information can be retrieved from RPC.

If `IsBan` is working - sure that event can be exposed - but not `.Player`

kantesilvestrs:
@Lauriichan `Connection of the player can also be pretty useful on join and leave, so in my opinion its pretty useful to have those informations` useful is a generous word here, because there is literally 0 information on `.Player` that might be considered completely unique, and definitely there is no information that you could link to outside sources for a permanent link.

Lauriichan:
> @Lauriichan Maybe instead of changing API interfaces you can look into to the actual events being generated and see if the Character information can be retrieved from RPC.
> 
> If `IsBan` is working - sure that event can be exposed - but not `.Player`

But for this I would need access to the main packages which aren't in the API package in Visual Studio

Lauriichan:
> @Lauriichan `Connection of the player can also be pretty useful on join and leave, so in my opinion its pretty useful to have those informations` useful is a generous word here, because there is literally 0 information on `.Player` that might be considered completely unique, and definitely there is no information that you could link to outside sources for a permanent link.

The Ip is enough since u need to login first on the system I'm working on

AeonLucid:
Please continue discussion on Discord

Lauriichan:
So the last thing I want to say is that if you don't want to merge it then close this I guess since I'm not here to discuss I just wanted to get the information I would need but if that is not possible then its okay to not merge but close it.
Thanks to all of you participating and discussing about this and for reviewing my pull request ^^

PR #216, opened 2020-12-07
Update readme to have instructions for ios

bill88t:
I just added some instructions on how to configure jailbroken ios to work with Impostor servers.
The tweak is not mine and I have nothing to do with it.
I tested it on my device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,14 @@ If you do not wish to execute any programs. Follow the instructions in [this web
 
 iOS devices need to be jailbroken in order to connect to Impostor servers. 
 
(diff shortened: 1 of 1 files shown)


gavinsteinhoff:
I think we didn't want to include any instructions on how to jailbreak an IOS device.

XtraCube:
He meant for people who already have a jailbroken device

bill88t:
> I think we didn't want to include any instructions on how to jailbreak an IOS device.

This does not include any instructions or refrences to how to jailbreak.
This purely instructs on how to configure an already jailbroken device to play on a impostor server.

PR #225, opened 2020-12-12
Update ClientManager.cs

SirOMGitsYOU:
Added Support for the latest Among Us version


js6pak:
Game version is not the same as protocol version, so 2020.12.9s **is** 2020.11.17

PR #234, opened 2020-12-16
Add announcement server

### LINKED ISSUE #186: [Feature Request] Announcement server implementation
# Feature Request
Enabling an Impostor server to serve announcements.
## Feature Information:
Announcement servers allow for sending info from the server to the users, and will be polled every time they go to the main menu. It supports [coloured text and links too](https://wiki.weewoo.net/wiki/Text), so it's not all plain text. All the information for setting up an announcement server should be found [here](https://wiki.weewoo.net/wiki/Announcements) Ideally an implementation would allow for servers to have a config option and a specific event to manage their announcements.

Hello World announcement test

js6pak:
### Description
- announcement server, can be disabled via config
- does nothing by default, provides plugin api
- terrible naming

### Closes issues

- Closes #186

### DIFF
--- a/src/Impostor.Api/Innersloth/Language.cs
+++ b/src/Impostor.Api/Innersloth/Language.cs
@@ -1,0 +1,11 @@
+namespace Impostor.Api.Innersloth
+{
+    public enum Language
+    {
+        English,
+        Spanish,
+        Portuguese,
+        Korean,
+        Russian,
+    }
+}
--- a/src/Impostor.Api/Net/Messages/Announcements/MessageHello.cs
+++ b/src/Impostor.Api/Net/Messages/Announcements/MessageHello.cs
@@ -1,0 +1,15 @@
+using Impostor.Api.Innersloth;
+
+namespace Impostor.Api.Net.Messages.Announcements
+{
+    public class MessageHello
+    {
+        public static void Deserialize(IMessageReader reader, out int announcementVersion, out int id, out Language language)
+        {
+            reader.ReadByte(); // TODO the hec is this
+            announcementVersion = reader.ReadPackedInt32();
+            id = reader.ReadPackedInt32();
+            language = (Language)reader.ReadPackedInt32();
+        }
+    }
+}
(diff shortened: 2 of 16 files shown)


src/Impostor.Api/Innersloth/Language.cs
```diff
@@ -0,0 +1,11 @@
+namespace Impostor.Api.Innersloth
+{
+    public enum Language
```
Mm2PL, line 3: Are these the same languages that are used for searching for games an such?

js6pak: Nope, you are talking about [GameKeywords](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Api/Innersloth/GameKeywords.cs), difference is that GameKeywords is only for room chat language while Language is whole among us translation (also different values for enums)
image

src/Impostor.Api/Net/Messages/Announcements/MessageHello.cs
```diff
@@ -0,0 +1,15 @@
+using Impostor.Api.Innersloth;
+
+namespace Impostor.Api.Net.Messages.Announcements
+{
+    public class MessageHello
+    {
+        public static void Deserialize(IMessageReader reader, out int announcementVersion, out int id, out Language language)
+        {
+            reader.ReadByte(); // TODO the hec is this
```
Sanae6, line 9: https://twitter.com/Sanae6k/status/1338631404758843392

js6pak: are you sure its a short? isn't it one byte? (first byte should be there because hazel)

Sanae6: both bytes are there because hazel
are you 100% sure it's actually the hazel version and not a stray byte?

Sanae6 (approved):

Sanae6:
is there anything that needs to be added before being merged? looks complete to me

XtraCube:
Bump

AlexejheroYTB (approved):
LGTM @AeonLucid

wolf20482:
seems good to me

PR #236, opened 2020-12-16
Refactor SetMurderedByAsync into MurderPlayerAsync

js6pak:
### Description

Refactor MurderPlayer so it makes sense

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
@@ -101,38 +101,30 @@ namespace Impostor.Server.Net.Inner.Objects
             await _game.FinishRpcAsync(writer, player.OwnerId);
         }
 
-        public async ValueTask SetMurderedByAsync(IClientPlayer impostor)
+        public async ValueTask MurderPlayerAsync(IInnerPlayerControl target)
         {
-            if (impostor.Character == null)
+            if (!PlayerInfo.IsImpostor)
             {
-                throw new ImpostorException("Character is null.");
-            }
-
-            if (!impostor.Character.PlayerInfo.IsImpostor)
-            {
-                throw new ImpostorProtocolException("Plugin tried to murder a player while the impostor specified was not an impostor.");
-            }
-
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
```diff
@@ -101,38 +101,30 @@ public async ValueTask SendChatToPlayerAsync(string text, IInnerPlayerControl? p
...
 
-            if (PlayerInfo.IsDead)
+            if (target.PlayerInfo.IsDead)
             {
-                return;
+                throw new ImpostorProtocolException("Tried to murder a player, but target was not alive.");
             }
 
-            // Update player.
             Die(DeathReason.Kill);
```
Mr-VIT, line 121:
```suggestion
            target.Die(DeathReason.Kill);
```

js6pak: whoops

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
```diff
@@ -101,38 +101,30 @@ public async ValueTask SendChatToPlayerAsync(string text, IInnerPlayerControl? p
...
-            if (PlayerInfo.IsDead)
+            if (target.PlayerInfo.IsDead)
             {
-                return;
+                throw new ImpostorProtocolException("Tried to murder a player, but target was not alive.");
             }
 
-            // Update player.
-            Die(DeathReason.Kill);
+            ((InnerPlayerControl)target).Die(DeathReason.Kill);
```
js6pak, line 121: how bad is this?

kantesilvestrs:
Can we just make sure that there is **'UGE** breaking changes notification added to the docs? I mean when this gets pushed - it will instantly break my plugin 😄 

Not to mention ==> people using docker will get the nightly update on next server restart and this will crash their servers, i.e. **me** <---

Do we really need to refactor existing function, can't we just add **deprecated** flag and add this method in parallel?

js6pak:
Api is experimental so I don't we should care about that yet, and this method is just awful right now

Issue #148, opened 2020-11-12
[API] Add missing events


AlexejheroYTB:
- [ ] Change color
- [ ] Change hat
- [ ] Change skin
- [ ] Change pet
- [ ] Start scanning
- [ ] Stop scanning
- [x] Enter vent
- [x] Exit vent
- [ ] Vote

PR #197, opened 2020-11-29
[API] Add Event: Player Activity Has Changed (currently only for started/stopped watching security camera)

Mr-VIT:
- [x] get player Ids started/stopped watching Security Camera
- [x] convert player Ids to IClientPlayer/IInnerPlayerControl
- [x] find a convenient(?) way to call event
- [x] upgrade to `PlayerActivityHasChangedEvent`

ToDo? :
- [ ] **await** `player.Controller.SetActivity` 
https://github.com/Impostor/Impostor/blob/bda1f1125ccf16563c6edae1010231c6b366107b/src/Impostor.Server/Net/Inner/Objects/Systems/ShipStatus/SecurityCameraSystemType.cs#L42

PR #237, opened 2020-12-16
Feature/add player vent event

Hiumee:
Addressing issue #148

Added event
+ `IPlayerVentEvent : IPlayerEvent`
  + `VentId : VentLocation`
  + `VentEnter : bool`

Added enum
+ `VentLocation : uint`

### DIFF
--- a/src/Impostor.Api/Innersloth/VentLocation.cs
+++ b/src/Impostor.Api/Innersloth/VentLocation.cs
@@ -1,0 +1,48 @@
+namespace Impostor.Api.Innersloth
+{
+    public enum VentLocation : uint
+    {
+        // Skeld
+        SkeldAdmin = 0,
+        SkeldRightHallway = 1,
+        SkeldCafeteria = 2,
+        SkeldElectrical = 3,
+        SkeldUpperEngine = 4,
+        SkeldSecurity = 5,
+        SkeldMedbay = 6,
+        SkeldWeapons = 7,
+        SkeldLowerReactor = 8,
+        SkeldLowerEngine = 9,
+        SkeldShields = 10,
+        SkeldUpperReactor = 11,
+        SkeldUpperNavigation = 12,
+        SkeldLowerNavigation = 13,
+
+        // Mira HQ
+        MiraBalcony = 1,
+        MiraCafeteria = 2,
+        MiraReactor = 3,
+        MiraLaboratory = 4,
+        MiraOffice = 5,
+        MiraAdmin = 6,
+        MiraGreenhouse = 7,
+        MiraMedbay = 8,
+        MiraDecontamination = 9,
+        MiraLockerRoom = 10,
+        MiraLaunchpad = 11,
+
+        // Polus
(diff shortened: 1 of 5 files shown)


AeonLucid:
Looks pretty good to me and I am okay with merging this, just not sure about the `VentLocation` enum. 
Wondering if anyone has ideas / thoughts about this before merging.

src/Impostor.Api/Innersloth/VentLocation.cs
```diff
@@ -0,0 +1,48 @@
...
+        SkeldWeapons = 7,
+        SkeldLowerReactor = 8,
+        SkeldLowerEngine = 9,
+        SkeldShields = 10,
+        SkeldUpperReactor = 11,
+        SkeldUpperNavigation = 12,
+        SkeldLowerNavigation = 13,
+
+        // Mira HQ
+        MiraBalcony = 1,
```
gavinsteinhoff, line 22: Why does this one start at 1?

src/Impostor.Api/Innersloth/VentLocation.cs
```diff
@@ -0,0 +1,48 @@
+namespace Impostor.Api.Innersloth
+{
+    public enum VentLocation : uint
```
gavinsteinhoff, line 3: Should these be separate enums? How does this work with them all being in the same enum?

AeonLucid: I think this is fine. If you write plugin code you can easily do 

```csharp
if (location == VentLocation.SkeldMedbay ||
    location == VentLocation.MiraMedbay ||
    location == VentLocation.PolusLaboratory) {

}
```

src/Impostor.Api/Innersloth/VentLocation.cs
Hiumee: > Why does this one start at 1?

Got the ids from here
https://github.com/codyphobe/among-us-protocol/blob/master/07_miscellaneous/04_map_specific_ids_for_vents_and_tasks.md

js6pak:
I feel like enum should be replace with something struct-like in Impostor.Api.Innersloth so it can include position - which would be required for anticheat

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -26,7 +26,7 @@ public InnerPlayerControl(ILogger<InnerPlayerControl> logger, IServiceProvider s
             _eventManager = eventManager;
             _game = game;
 
-            Physics = ActivatorUtilities.CreateInstance<InnerPlayerPhysics>(serviceProvider, this);
+            Physics = ActivatorUtilities.CreateInstance<InnerPlayerPhysics>(serviceProvider, this, _eventManager, _game);
```
Mr-VIT, line 29: https://github.com/Impostor/Impostor/pull/197#discussion_r532611352
> Why is _eventManager here? It is in the serviceProvider.

PR #239, opened 2020-12-16
 Game privacy

### LINKED ISSUE #150: [API] Add set privacy api function
Use case: Forcing a server to be public?

ZiadMontaser:
### Description

add SetGamePrivacyAsync

### Closes issues

- closes #150

### DIFF
--- a/src/Impostor.Api/Innersloth/GamePrivacy.cs
+++ b/src/Impostor.Api/Innersloth/GamePrivacy.cs
@@ -1,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Impostor.Api.Innersloth
+{
+    public enum GamePrivacy
+    {
+        Private,
+        Public,
+    }
+}
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 Impostor is one of the first **Among Us** private servers, written in C#. 
 
-We support steam, itch, android and iOS. The latest version supported is `2020.9.22`, the `dev` build currently supports `2020.10.22`.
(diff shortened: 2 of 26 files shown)


js6pak:
What is wrong with this commit history

src/Impostor.Api/Innersloth/GamePrivacy.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Impostor.Api.Innersloth
+{
+    public enum GamePrivacy
```
js6pak, line 7: Can't this be just a bool?

ZiadMontaser: yes but i think this makes more sense

Minorusama: I also think it should be a bool, since all you do with this enum is
`IsPublic = privace == GamePrivacy.Public;`

IDjinn: why don't make it just `Default `, `Private` or `Public`? so, default is configured in settings, like always private or always public?

Minorusama: > why don't make it just `Default`, `Private` or `Public`? so, default is configured in settings, like always private or always public?

That doesn't make sense. The point of the method is to set the privacy of the room by plugin. What would `Default` mean ?
Having a setting in `config.json` to set the default privacy could be added, but I don't think it would be useful.

PR #262, opened 2020-12-28
Documentation: Fix plugin path for loading libraries

jrutley:
### Description

The "libraries" path in the configuration doesn't match the config in PluginConfig.cs, so this change lines those up


ZiadMontaser:
duplicate of #223

PR #266, opened 2021-01-02
BeforeGameCreatedEvent

Minorusama:
### Description

Adds a `BeforeGameCreatedEvent`. The event contains a method `TryToSetCode` to allow plugins to choose the gamecode. It returns true if the code is available and false if it already exists. It also throws an `ArgumentException` if the argument is not a valid gamecode.


src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,50 @@
...
+        private readonly ICollection<int> _gameKeys;
+
+        public BeforeGameCreatedEvent(ICollection<int> gameKeys)
+        {
+            _gameKeys = gameKeys;
+        }
+
+        public string? GameCode { get; private set; }
+
+        public bool TryToSetCode(string gameCode)
```
js6pak, line 24: just use property setter with backing field

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class BeforeGameCreatedEvent : IBeforeGameCreatedEvent
+    {
+        private readonly ICollection<int> _gameKeys;
+
+        public BeforeGameCreatedEvent(ICollection<int> gameKeys)
+        {
+            _gameKeys = gameKeys;
+        }
+
+        public string? GameCode { get; private set; }
```
js6pak, line 22: use [GameCode](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Api/Games/GameCode.cs)

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public bool TryToSetCode(string gameCode)
+        {
+            if (gameCode == null)
```
js6pak, line 26: already implemented in [GameCodeParser](https://github.com/Impostor/Impostor/blob/dev/src/Impostor.Api/Innersloth/GameCodeParser.cs)?

src/Impostor.Server/Net/Manager/GameManager.cs
```diff
@@ -62,9 +65,9 @@ public async ValueTask<IGame> CreateAsync(GameOptionsData options)
             return game;
         }
 
-        private async ValueTask<(bool success, Game game)> TryCreateAsync(GameOptionsData options)
+        private async ValueTask<(bool success, Game game)> TryCreateAsync(GameOptionsData options, string? desiredGameCode = null)
         {
-            var gameCode = _gameCodeFactory.Create();
+            GameCode gameCode = desiredGameCode ?? _gameCodeFactory.Create();
```
js6pak, line 70: follow code style - use `var`

Minorusama:
Thanks for the review, I'll just answer all 4 comments here :

1. The point of `TryToSetCode` is to return true if the code is available. That way, if the first code you want already exists, you can use another one, and so on. That would be impossible with a property setter.
2. At first, I didn't want to instantiate a `GameCode` but I guess you're right. I'll change that.
3. I didn't see anything like that in `GameCodeParser` 🤔
4. Ok

src/Impostor.Api/Events/Game/IBeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// <summary>
+        ///     Gets the <see cref="Games.GameCode"/> set by this event.
+        /// </summary>
+        GameCode? GameCode { get; }
+
+        /// <summary>
+        ///     Tries to set the game code to be used. This doesn't garantee the code will be used, if another lobby has been created with the same code before this event returns for exemple.
+        /// </summary>
+        /// <param name="gameCode">The code to try to set.</param>
+        /// <exception cref="ArgumentException">If the argument is not 6 characters long or does not contains only letters.</exception>
```
js6pak, line 19: `ArgumentException` -> `System.ArgumentException`

src/Impostor.Server/Net/Manager/GameManager.cs
js6pak: > Ok

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
js6pak: > I didn't see anything like that in `GameCodeParser` :thinking:

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
js6pak: > At first, I didn't want to instantiate a `GameCode` but I guess you're right. I'll change that.

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
js6pak: > The point of `TryToSetCode` is to return true if the code is available. That way, if the first code you want already exists, you can use another one, and so on. That would be impossible with a property setter.

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+using Impostor.Api.Games;
+using Impostor.Api.Innersloth;
+using Impostor.Server.Net.Redirector;
+using Impostor.Server.Net.State;
+
+namespace Impostor.Server.Events
+{
+    public class BeforeGameCreatedEvent : IBeforeGameCreatedEvent
+    {
+        private readonly ICollection<int> _gameKeys;
```
js6pak, line 16: imo `GameManager` reference should be passed, and then instead of `ICollection.Contains` just use `GameManager.Find` and check for null

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+        public BeforeGameCreatedEvent(ICollection<int> gameKeys)
+        {
+            _gameKeys = gameKeys;
+        }
+
+        public GameCode? GameCode { get; private set; }
+
+        public bool TryToSetCode(GameCode gameCode)
+        {
+            if (gameCode.Code.Length != 6)
```
js6pak, line 27: What if `GameCode` is v1 (or better, v3?). And code for it is already present in [`GameCodeParser`](https://github.com/Impostor/Impostor/blob/dc759b1d573b59ff5c39dd2792c40ebdec7649a2/src/Impostor.Api/Innersloth/GameCodeParser.cs#L95) anyways, just check if value is -1, or better, make a new property in GameCode which is just `IsInvalid => Value == -1`

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                throw new ArgumentException("GameCode must be 6 characters longs.");
+            }
+
+            if (!Regex.IsMatch(gameCode.Code, @"^[A-Z]+$"))
```
js6pak, line 32: Again, this would never return false, it would just crash in https://github.com/Impostor/Impostor/blob/dc759b1d573b59ff5c39dd2792c40ebdec7649a2/src/Impostor.Api/Innersloth/GameCodeParser.cs#L98-L111

js6pak (changes requested):

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
js6pak: It still could be a property, just throw a exception that you can catch, no?

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
js6pak: Explained better in new comments

src/Impostor.Server/Events/Game/BeforeGameCreatedEvent.cs
Minorusama: So the plugin should catch the exception ? I find it a little dangerous, as unhandled exception would completely prevent the room creation. I know you could say that it would be "the plugin's fault" but still.

js6pak: It wouldn't actually, exception would be thrown inside event handling - so just error would be printed out

Minorusama: Well ok then.

js6pak (approved):

PR #238, opened 2020-12-16
Replace ColorId, HatId, PetId, SkinId with their respective enums

js6pak:
### Description

Replace character customization ids with enums to reduce redundancy. Enums are compiled to numbers (type is specified after enum name) at runtime, so you can just use 
```cs
IInnerPlayerControl.SetSkinAsync((SkinType) 123);
``` 
anyway.

### Closes issues

- closes #161

PR #267, opened 2021-01-03
Add Geoff Keighley Mask

skick1234:
### Description

Add Geoff Keighley Mask to HatType.

Available on 2020.12.9 (or above)


Minorusama:
Thanks for the PR but this is already covered by #238

PR #275, opened 2021-01-07
Add a property to identify a task and methods to complete tasks

### LINKED ISSUE #202: TaskType currently not known for tasks
Since https://github.com/Impostor/Impostor/pull/164 fixed the index being used as TaskTypes, we don't have any information about the TaskTypes in memory now.

The `TaskInfo.Type` property should be properly updated to include the proper `TaskTypes`.

### LINKED ISSUE #261: Complete Task Suggestion
# Api Suggestion
## Suggestion
Allow completing tasks from specific players. (currently it's `get;` only)  
eg.: `IInnerPlayerInfo#Tasks.ToList()[0].SetCompleteAsync(bool);`

## Use cases
• Programmatically complete tasks for a single player;  
• Simulate a timer in the taskbar.

Minorusama:
### Description

- Adds a method to complete a task in TaskInfo
- Adds a method to complete all tasks in InnerPlayerControl
- Adds a TaskTypes in TaskInfo to identify the task

### Closes issues

- closes #261
- closes #202


Elsensee:
I think it might be beneficial to have TaskCategories be a flag enum, so we can assume both VisualTask and Short/LongTask to the same task?

Minorusama:
> I think it might be beneficial to have TaskCategories be a flag enum, so we can assume both VisualTask and Short/LongTask to the same task?

I thought about doing so but decided not to, because I figured the two informations (Common/Short/Long and Visual) have nothing to do with each other and it will make using switch on TaskCategories easier.

That being said, I'm not against doing what you suggest if other people think the same.

js6pak:
I really don't like the idea of using enums for tasks, vents etc. Just make it a class/struct and make per map list of it

js6pak:
This PR would have to be updated for airship and follow new static map data strucutre (https://github.com/Impostor/Impostor/tree/master/src/Impostor.Api/Innersloth/Maps). I might do it later.

Minorusama:
The rework has been submitted in #395. I'm closing this one.

PR #279, opened 2021-01-09
Really fix SetInfectedAsync's packet structure

libhalt:
**Description**

New Impostor list is prefixed with the list's length not the host's netid

**Use Case & Test**
https://gist.github.com/libhalt/e3f98c717a29514dcc259c6ca9c7f708

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Api.cs
+++ b/src/Impostor.Server/Net/State/Game.Api.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using Impostor.Api;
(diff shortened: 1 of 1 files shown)


js6pak:
Even if the packet is now correct, are you sure your use case is even working? (without any side effects)

libhalt:
The above test plugin has been tested locally on itch version of the game with 5 client running locally.
(the game end screen showed the correct impostor, only the impostor had red name tag , etc)

Elsensee:
He meant, it's usually not possible to set the impostors without client-side modifications.
Either the host would be de-synced or the clients would ignore the change (depending on _when_ you send the SetInfectedRPC)

libhalt:
I agree with your/his statement that this only works at a very specific situation but it is usefull at that very spefic situation. Initially, i thought about limiting the call to during IGameStartingEvent but that is a "breaking change" that would require edit of the doc (although i cant imagine any plugin using this method outside the limit that will be imposed as it would be non functional) and outside the scope of this PR. However, if the maintainers wish this method to be limited I can add that in

PR #274, opened 2021-01-07
Anticheat disable config

20PercentRendered:
### Description
PR to add config option to disable anticheat

### Closes issues

- closes #263

PR #296, opened 2021-01-22
Added anticheat config

### LINKED ISSUE #263: Ability to disable server-side anticheat via config.json
# Feature Request
Allow hosts to disable server-side anticheat via config.json
## Feature Information:

Disables server-side anticheat with a true/false boolean in config.json

## I confirm:
- [x] that I have searched for an existing feature request matching the description.

XtraCube:
### Closes issues

- closes #263

### DIFF
--- a/src/Impostor.Server/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Server/Config/AntiCheatConfig.cs
@@ -1,9 +1,13 @@
-﻿namespace Impostor.Server.Config
+namespace Impostor.Server.Config
 {
     public class AntiCheatConfig
     {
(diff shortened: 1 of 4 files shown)


Elsensee:
You're change only removes the kicking after a cheat was detected. So, this is even worse behaviour since the packets won't be sent to the other players but they'll stop at the server. You need to prevent every throw before it happens, since it'll disrupt the program flow.

AeonLucid:
This is not the way, sorry.

js6pak:
Wtf is with these spaces

js6pak:
And its duplicate of #117 and #274

PR #331, opened 2021-03-03
Consolidate warning configuration into global file

miniduikboot:
### Description

Currently you get quite a lot of warnings when building Impostor, 176 to be exact. Most of them are harmless, but some of them are possibly quite bad. Furthermore, we don't want this number to go up any higher, because that can start masking new bugs. This commit is radical: every warning that's currently thrown is ignored project-wide, reducing the warning count to 4.

There were also two identical Stylecop rulesets, one for Api, one for Server, make them global and annotate which rules are disabled (comments from the default SC rules)

Followup pull requests should start removing these items category for category:

1. There are 4 Nuget problems related to Impostor.Tools.Proxy using a package that targets .NET Framework instead of .NET Core.
2. Nullability problems: There are 67 instances of these problems (CS8xxx)
3. Possible bigger issues like `src/Impostor.Hazel/Udp/UdpServerConnection.cs(91,17): warning CS4014: Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call`. Please look through the warnings list
4. The remainder is mostly Stylecop warnings: we should decide how much of them we care about these and fix them.


js6pak:
Implemented in #333

PR #338, opened 2021-03-05
Send GameOptionData and player customization (Name, Color, etc...) to a specific player

Minorusama:
Allow to desync GameOptionData and player customization.
Things you can do :
- Set the name, color, hat, skin and pet of a player to be only visible by them or another player
- Change the GameOptionData for one player

Gameplay ideas :
- Give information during gameplay to a player by changing their name only to them
- When comms are sabotaged, all crewmate see everyone with the same color/skin and no name, but impostors are not affected
- Change the vision range of a player based on their completed task

### DIFF
--- a/src/Impostor.Api/Games/IGame.cs
+++ b/src/Impostor.Api/Games/IGame.cs
@@ -1,9 +1,10 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
 using Impostor.Api.Innersloth;
 using Impostor.Api.Net;
 using Impostor.Api.Net.Inner;
+using Impostor.Api.Net.Inner.Objects;
 using Impostor.Api.Net.Messages;
 
 namespace Impostor.Api.Games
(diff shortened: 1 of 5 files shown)


js6pak:
This is a big design choice, imo base plugin api shouldn't be able to cause desyncs, but it should be possible to make them by sending rpcs yourself.

Minorusama:
> This is a big design choice, imo base plugin api shouldn't be able to cause desyncs, but it should be possible to make them by sending rpcs yourself.

Can you explain why you think so ?

Personally, I think allowing desyncs is a requested feature and opens a lot of possibilities for plugins. I know you could do it with custom rpc, but this way is so much simpler.

IrishBruse:
its not difficult to make a custom class to handle this stuff but why not add them as extension methods and have them in a 
`using Imposter.experimental.rpcDesyncs`

js6pak:
I agree, but it could very well be a plugin

Galster-dev:
> 
> 
> I agree, but it could very well be a plugin

But plugin dependencies are not implemented, are they?

IrishBruse:
You dont actually need that for you to use other plugins i have a plugin that uses another one for chat commands

HarmHoog:
@Galster-dev Not if you're plugin needs to use dependencies at startup. I actually created an API plugin that adds some desync extension methods to the player class (https://github.com/HarmHoog/ExtraPlayerFunctions); as long as you don't call any methods too early it will work fine as a dependency.

IrishBruse:
I made something similar to that but mine is just a plain dll and i pass it the writer provider in a constructor

Fluffy728:
Oh no it has conflicts that must be resolved!

Fluffy728:
It is a big design choice anyway

Fluffy728:
Sorry I necroposted

Fluffy728:
Yeah and also I reliized it is fixed

PR #371, opened 2021-04-01
Add proper Linux commands for Docs

JantsoP:
### Description

This PR will add proper commands to run Impostor server on linux.
You cannot just download the .zip, unzip it and run it. You must set permissions first.
Otherwise, you will just get errors from everywhere. Found out that these permissions hit the sweet spot for running Impostor on Debian/Ubuntu.

### DIFF
--- a/docs/Running-the-server.md
+++ b/docs/Running-the-server.md
@@ -13,7 +13,7 @@ There are currently two modes to run the Impostor server in. The first way is th
 3. Download either the Windows or the Linux version.
 4. Extract the zip.
 5. Modify `config.json` to your liking. Documentation can be found [here](Server-configuration.md) *(this step is mandatory if you want to expose this server to other devices)*
-6. LINUX ONLY! Set permissions 755 to Impostor.Server / plugins / libraries with chmod 755 Impostor.Server plugins libraries. Set 644 to config.json with chmod 644 config.json
+6. LINUX ONLY! Set permissions 755 to Impostor.Server / plugins / libraries with `chmod 755 Impostor.Server plugins libraries`. Set 644 to config.json with `chmod 644 config.json`.
 7. Run `Impostor.Server.exe` (Windows) / `Impostor.Server` (Linux)
 
 ### Using docker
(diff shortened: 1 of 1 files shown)


JantsoP:
2fa3bcb commit will also put proper command to run Impostor.Server on headless machines

js6pak:
Both permissions and way you are launching the server are user specific, you should handle it on your own

JantsoP:
With build 159 and unzip, all files will come w/o any permissions.
 As seen here: 
E967td4U

Hence proper permissions must be set before running.

miniduikboot:
Would it be possible to change the CI build instead so it can set the proper permissions?

dimaguy:
Permissions are not easy to get right, you can't possibly cover every case with those 2 commands (by the way chmod +x does a better job), especially when you don't explain them.  
You need to take into account that there will be people wgetting stuff with sudo or put stuff in a restricted folder and try opening it with regular user.  
IIRC You dont need executable bit on the libraries nor plugins, as they are executed by the server which has the executable bit already, you only need to make them readable to the user executing the server
These kinds of stuff should be left for the user to figure out, as it is an important thing to understand about the system.

JantsoP:
I ran chmod +x and it did not work. After setting 755 it started working but I got errors about libraries not found and same with plugins. You guys should reproduce the issue before even starting to say what to do and not what to do.

dimaguy:
> 
> 
> I ran chmod +x and it did not work. After setting 755 it started working but I got errors about libraries not found and same with plugins. You guys should reproduce the issue before even starting to say what to do and not what to do.

Heres where I ask you to provide an ls -la of the entire thing.
But not here, since that's out of the scope of this PR, go on discord.

JantsoP:
part1
part2
part3


Here you go. Just did this on my Debian 10 machine. (Address already in use you can ignore. I know that because already fixed the issues last night)

dimaguy:
Again ON DISCORD, this is out of scope for this PR

dimaguy:
By the way, as proof that not every case is the same:   
imagem  
This is a freshly wgetted&unzipped Impostor Server, it automatically sets executable bit. and I haven't even had to use chmod

js6pak:
Fixed in 0a30fe5a3e69a242112875b677b109fc838dc0f4

PR #372, opened 2021-04-02
Protocol changes for 2021.3.31

miniduikboot:
### Description

This PR makes most of the major changes that enable 2021.3.31 clients to connect to Impostor

### Closes issues

- closes #368

PR #375, opened 2021-04-04
Allow Android Versions

JohnTranQUT:
### Description

Allow android phones onto airship impostor server

### DIFF
--- a/src/Impostor.Server/Net/Manager/ClientManager.cs
+++ b/src/Impostor.Server/Net/Manager/ClientManager.cs
@@ -20,7 +20,7 @@ namespace Impostor.Server.Net.Manager
         private static readonly HashSet<int> SupportedVersions = new HashSet<int>
         {
             GameVersion.GetVersion(2021, 3, 25), // 2021.3.31s
-            GameVersion.GetVersion(2021, 4, 2), // 2021.4.2a andriod
+            GameVersion.GetVersion(2021, 4, 2), // 2021.4.2a
         };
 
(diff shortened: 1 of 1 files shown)


js6pak:
If forte decided to change the protocol version, are you sure it will work?

JohnTranQUT:
> If forte decided to change the protocol version, are you sure it will work?

I had not done anything to the protocol however I tested a few games with andriod/bluestacks and it appears to be running smoothly with the https://github.com/Impostor/Impostor/pull/372 .

I will report back as soon there are any concerns.

src/Impostor.Server/Net/Manager/ClientManager.cs
```diff
@@ -20,6 +20,7 @@ internal partial class ClientManager
         private static readonly HashSet<int> SupportedVersions = new HashSet<int>
         {
             GameVersion.GetVersion(2021, 3, 25), // 2021.3.31s
+            GameVersion.GetVersion(2021, 4, 2), // 2021.4.2a andriod
```
js6pak, line 23:
```suggestion
            GameVersion.GetVersion(2021, 4, 2), // 2021.4.2a
```

js6pak (changes requested):

src/Impostor.Server/Net/Manager/ClientManager.cs
JohnTranQUT: Yep, spelt it wrong. My bad.

js6pak (approved):

PR #376, opened 2021-04-04
Fix SnapTo on Airship spawn triggering Anticheat

miniduikboot:
### Description

As part of initial spawn, Among Us will call a SnapTo to -25,40 which is
an out of bounds position.

### Closes issues

- Starting a game on Airship with AC enabled kicks every crewmate before they spawn

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.cs
@@ -1,7 +1,9 @@
-﻿using System.Numerics;
+using System;
+using System.Numerics;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Impostor.Server/Net/Inner/Objects/Components/InnerCustomNetworkTransform.cs
```diff
@@ -22,6 +24,8 @@ internal partial class InnerCustomNetworkTransform : InnerNetObject
 
         private ushort _lastSequenceId;
 
+        private bool _airshipAllowSnapTo = true;
+
```
kantesilvestrs, line 28: Should this be name something more generic and not specific to a map?

miniduikboot:
Current version is flawed, will rework

PR #385, opened 2021-04-08
Allow plugins to set dependencies, softdependencies and loadbefores to determine the load order of plugins

HarmHoog:
### Description
This PR adds the option for plugin developers to set the name of plugins they depend on to ensure the plugins they depend on are loaded. In the ImpostorPluginAttribute class they can now add:

- Dependencies: An array containing the names of hard dependencies can be provided. If one of these hard dependencies is missing, the server will log an error with the missing dependency name.
- Softdependencies: An array containing the names of soft dependencies. Just like the dependencies they will load before the plugin but will not give an error if the soft dependency is missing.
- Loadbefore: An array containing the names of plugins that have to be loaded after the current plugin is loaded.


realfaceitkantesilvestrs:
@HarmHoog good idea, but this is really needs config to finally be in it's own folder, otherwise this is not container friendly, i.e. I don't want to mount the whole container so that I could just change the config file. Or at least allow config file extension?

js6pak:
I think that api like
```cs
    [ImpostorPlugin("gg.impostor.example", "Example", "AeonLucid", "1.0.0")]
    [ImpostorDependency("example.hardapi", DependencyType.Hard)]
	[ImpostorDependency("example.softapi", DependencyType.Soft)]
	[ImpostorDependency("example.plugin", DependencyType.LoadBefore)]
    public class ExamplePlugin : PluginBase
    {
		...
    }
```
would look much better.

HarmHoog:
@js6pak I'll look into that way next week when I've some time, I guess the dependencies will then be added as attributes to the assembly? I was also thinking about what would be the right way to handle a missing hard dependency. Right now it throws some error but still attempts to start the plugin without it. It could also be made so that a missing hard dependency will block the plugin from starting.

js6pak:
>  I guess the dependencies will then be added as attributes to the assembly

they would work just like ImpostorPluginAttribute works currently, just add it on plugin class

> I was also thinking about what would be the right way to handle a missing hard dependency

Hard dependency is missing, plugin should not load at all
Soft dependency is here, load plugin after that dependency, if missing, do nothing
Load before dependency is here, load plugin before that dependency, if missing, do nothing

HarmHoog:
@js6pak Unexpectedly had some time today so I implemented this. It will now give an error message and disable a plugin when a hard dependency is not present. The dependencies can now be defined as attributes in the PluginBase class.

src/Impostor.Server/Plugins/PluginLoader.cs
```diff
@@ -155,5 +167,103 @@ private static void CheckPaths(IEnumerable<string> paths)
...
+                    if (dependency.DependencyType == DependencyType.LoadBefore)
+                    {
+                        dependencyGraph[dependency.Name].Add(plugin);
+                    }
+                    else
+                    {
+                        dependencyGraph[plugin].Add(dependency.Name);
+                    }
+                }
+            }
```
realfaceitkantesilvestrs, line 206: Any checks when there is a circular dependency?

HarmHoog: No, right now if there is a circular dependency the server will arbitrarily load either one of the plugins first. A circularity check could be added although I'm not sure whether it is really needed since if there exists a circular dependency the plugin is just not written correctly.

PR #389, opened 2021-04-13
Make it possible to read plugin information from csproj attributes

### LINKED ISSUE #172: Read plugin info from Assembly Info instead of ImpostorPluginAttribute arguments
Replacement table:
|Current argument|Assembly attribute|.csproj property
|-:|-|-
|name|Title|AssemblyTitle
|author|Company|Authors
|version|InformationalVersion|Version

dev writes into `Impostor.Plugins.Example.csproj`:
```csproj
<Project Sdk="Microsoft.NET.Sdk">
...
  <PropertyGroup>
    <TargetFramework>netstandard2.1</TargetFramework>

    <Version>1.2.3</Version>
    <AssemblyTitle>Example</AssemblyTitle>
    <Authors>AeonLucid</Authors>
  </PropertyGroup>
...
</Project>
```

js6pak:
Closes #172

### DIFF
--- a/src/Impostor.Api/Plugins/ImpostorPluginAttribute.cs
+++ b/src/Impostor.Api/Plugins/ImpostorPluginAttribute.cs
@@ -10,6 +10,7 @@ namespace Impostor.Api.Plugins
             Id = id;
         }
 
+        [Obsolete("Use (string id) constructor to avoid redundancy")]
         public ImpostorPluginAttribute(string id, string name, string author, string version)
         {
             Id = id;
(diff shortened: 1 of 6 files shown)


HarmHoog:
Isn't Id actually not somewhat unnecessary when there is a Name attribute?

js6pak:
Name is meant to be just a human readable string, id is something that has to be always unique

Mr-VIT (approved):

HarmHoog:
@js6pak I get that. Though it has to be unique within a list of active plugins on a server right? I don't think it's likely that a server has two plugins with exactly the same name installed so that's why I would argue that an Id is unnecessary or double.

PR #404, opened 2021-04-28
Add custom message support

js6pak:


src/Impostor.Api/Games/IGame.cs
```diff
@@ -90,5 +90,24 @@ public interface IGame
...
         /// <returns>A <see cref="ValueTask" /> representing the asynchronous operation.</returns>
         ValueTask SendToAsync(IMessageWriter writer, int id);
+
+        /// <summary>
+        ///     Creates a MessageWriter with GameData header.
+        /// </summary>
+        /// <remarks>Use with caution, recommended only for advanced developers.</remarks>
+        /// <param name="targetNetId">Net id of the InnerNetObject.</param>
+        /// <param name="callId">Rpc id of the message.</param>
+        /// <param name="targetClientId">Client id of GameDataTo's target.</param>
```
miniduikboot, line 100: maybe document what the value `-1` does

src/Impostor.Api/Utils/IDateTimeProvider.cs
```diff
@@ -1,6 +1,6 @@
 using System;
 
-namespace Impostor.Api
+namespace Impostor.Api.Utils
```
miniduikboot, line 3: why move classes from the public api? they were stabilized in this location, why do they need to be moved? this is pointless breakage

miniduikboot (commented):
looked through it, didn't compile or run the code

src/Impostor.Api/Games/IGame.cs
js6pak: Or just make it a nullable int?

src/Impostor.Api/Utils/IDateTimeProvider.cs
js6pak: To keep things clean, it really wasn't supposed to be in root namespace, and no one used it anyway

src/Impostor.Api/Games/IGame.cs
miniduikboot: that's probably even better, but document it

PR #408, opened 2021-05-10
Site: Add default regions to region file

miniduikboot:
### Description
In the latest AU patch the menu is now generated based on this file. To
allow users to switch back to the default servers, add them to the file
as well.

### Closes issues

- server list is empty except for the single custom server


js6pak:
I thought about making it more like a editor, as you can have more than 1 server now

PR #423, opened 2021-06-15
2021.6.15 support

miniduikboot:
### Description

This PR adds support for version 2021.6.15 which just released, bringing with it 15 player support, new colors and long awaited bugfixes like the toilet doirs in Airship not being networked properly.

ToDo on this PR:

- [x] The MeetingHud was reworked by Innersloth, this is not included yet in this PR. As the new meeting messages are longer than the old ones, games will appear to play fine, but the recently merged Voting API will not yet work properly.
- [ ] The new QuickChat RPC can use an API that exposes it to plugin developers.
- [ ] Testing has been minimal so far, help is appreciated. I updated my public test server, `duikbo.at`. False AC violations also welcome.
- [x] Readme update to state this new version. As this will likely need updates once Innersloth starts pushing hotfixes, I'll push this off for a little while.

### Closes issues

- nobody opened an issue yet to ask about 2021.6.15 support :smile: 
- closes #422


src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -565,6 +577,12 @@ private async ValueTask<bool> HandleSendChat(ClientPlayer sender, string message
             return !@event.IsCancelled;
         }
 
+        private ValueTask<bool> HandleSendQuickChat(ClientPlayer sender, IQuickChatMessage message)
+        {
+            // TODO add event
+            return new ValueTask<bool>(true);
+        }
```
realfaceitkantesilvestrs, line 584: Can we make quick chat cancellable event so that we could use this for some custom things in plugins?

src/Impostor.Server/Net/Inner/Objects/ShipStatus/InnerPolusShipStatus.cs
```diff
@@ -36,13 +36,15 @@ public override Vector2 GetSpawnLocation(InnerPlayerControl player, int numPlaye
             }
 
             Vector2 position;
-            if (player.PlayerId < 5)
+            int halfPlayers = numPlayers / 2; // floored intentionally
+            int spawnId = player.PlayerId % 15;
```
realfaceitkantesilvestrs, line 40: Is this not going to have the same issues voting had with NetId? Where if you disconnect and connect to the lobby your PlayerId/netId and order do not match anymore?

src/Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>1.4.0</VersionPrefix>
+    <VersionPrefix>1.5.0</VersionPrefix>
```
realfaceitkantesilvestrs, line 3: Looking at the last few releases it seems more logical to match server version with the game version, but that's for future

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: Good idea, noted.

src/Directory.Build.props
miniduikboot: I'm not sure what you mean with this remark. While it might be useful to link release numbering to game versions, marketing versions and network broadcast versions rarely match each other. I don't think it is useful to change Impostor's versioning to match either:
-  If we match the marketing versions (what's on the main menu) that would mean that we must bump for every patch release including those that don't have or the latter and that an Impostor version would be compatible with versions even those that don't involve network changes
- If we match the network version (which is what the game identifies itself as), we have the issue that network versions don't clearly match marketing versions. For example, Marketing version 2021.6.15 uses Network version 2021.4.25, which is confusing as marketing version 2021.5.25.2 also exist, which uses a network version of 2021.4.2.

Current versioning scheme isn't perfect, but I haven't heard a good suggestion for a new one yet

realfaceitkantesilvestrs: Fair enough, I see the issue.

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.PlayerVoteArea.cs
```diff
@@ -34,34 +29,34 @@ public PlayerVoteArea(InnerMeetingHud parent, InnerPlayerInfo targetPlayer, bool
...
+                            VoteType = (VoteType)value;
+                            break;
+
+                        default:
+                            VoteType = TVoteType.Player;
+                            VotedFor = Parent.Game.GameNet.GameData!.GetPlayerById(value)?.Controller;
+                            break;
                     }
+
+                    DidVote = VoteType != TVoteType.HasNotVoted;
```
miniduikboot, line 59: why store it as a separate property instead of implementing a getter that does the compare then?

idem for IsDead

SakuMai:
Can I download a server that already supports the new version of among as?

miniduikboot:
Now that the Voting API is updated by 6pak, I suggest to review and merge this PR.

- MeetingHud was updated by 6pak, the voting API should work again. Please test.
- QuickChat API will be backed out and placed in a different PR. It needs some time to be fully implemented and I don't want to block merging this PR.
- Testing so far did not reveal issues, so I want to make this available for wider testing
- Readme has been updated to include new install instructions and to remove the table.

PR #432, opened 2021-07-13
Improve error when client version is unsupported

miniduikboot:
### Description

In the default message that was used previously the user is asked to
upgrade his game. This is not always correct, as sometimes the Impostor
server is too old and should be updated instead. Therefore custom
disconnect messages are used that blames either the user or the server
operator, which is hopefully more accurate.

I've chosen not to include the game version that is supported by the
server in this message because the network version rarely matches the
advertised version in the main menu.

### Closes issues

- prevents many issues of people trying to set up Impostor for the first time and downloading the wrong version

### DIFF
--- a/src/Impostor.Server/Net/Manager/ClientManager.cs
+++ b/src/Impostor.Server/Net/Manager/ClientManager.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Concurrent;
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/Manager/ClientManager.cs
```diff
@@ -57,13 +66,26 @@ public int NextId()
...
+                        Message01JoinGameS2C.SerializeError(packet, false, DisconnectReason.Custom, DisconnectMessages.VersionClientTooOld);
+                        break;
+                    case VersionCompareResult.ServerTooOld:
+                        Message01JoinGameS2C.SerializeError(packet, false, DisconnectReason.Custom, DisconnectMessages.VersionServerTooOld);
+                        break;
+                    case VersionCompareResult.Unknown:
+                        Message01JoinGameS2C.SerializeError(packet, false, DisconnectReason.Custom, DisconnectMessages.VersionUnsupported);
+                        break;
+                }
+
```
js6pak, line 88:
```suggestion
                var message = versionCompare switch
                {
                    VersionCompareResult.ClientTooOld => DisconnectMessages.VersionClientTooOld,
                    VersionCompareResult.ServerTooOld => DisconnectMessages.VersionServerTooOld,
                    VersionCompareResult.Unknown => DisconnectMessages.VersionUnsupported,
                    _ => throw new ArgumentOutOfRangeException(),
                };

                Message01JoinGameS2C.SerializeError(packet, false, DisconnectReason.Custom, message);

```

js6pak (changes requested):

js6pak (approved):

PR #442, opened 2021-11-30
2021.11.9 and 2021.12.14 support

### LINKED ISSUE #440: Possibly broken with AU v2021.11.9s
## Base Information
- Impostor v1.6.0 on Ubuntu 18.04.5 LTS
- Among Us v2021.11.9s on Windows 10

## I confirm:
- [x] that I have searched for an existing bug report for this issue.

## Symptoms

After the recent AU update, it can no longer host or find games on my Impostor instance. The error is "Reliable packet 1 (size=66) was not ack'd after 7503ms (9 resends)." Repeated attempts occasionally produce "An address incompatible with the requested protocol was used" instead. There is no indication of a connection attempt in the server-side Impostor output. I cannot find a client-side log to check.

I watched tcpdump while attempting to host a game and discovered something strange. No packets arrive on the configured UDP port; instead they arrive on port N+3, where N is whatever port is configured in the client's `regionInfo.json`. Reducing the port in regionInfo.json by 3 causes packets to arrive on the expected port according to tcpdump, but does not change the in-game symptoms.

I can give you the address of my server privately if it will help.

## Reproduction

1. Start an Impostor server
2. Configure its address in `regionInfo.json`
3. (optional): On the server, run `tcpdump udp and host $client_ip` to observe the connection.
3. Start AU, select the private server, and try to host a game.

miniduikboot:
### Description

This PR updates Impostor to handle the protocol changes introduced by v2021.11.9

Technical changelog:
- Announcement server is dead
- Auth server is also dead
- Customization RPC's are replaced by new ones that accept a string instead of an ID
- PlatformSpecificData was thrown in where needed
- New roles were implemented. Not yet exposed to plugin API.

Known issues:
- It is impossible to connect to a custom server due to encryption being force enabled on all unmodded clients. Patches have been created for [Reactor](https://github.com/NuclearPowered/Reactor), I have isolated these patches into a [standalone mod](https://github.com/miniduikboot/Mini.RegionPortFixer). Innersloth has been contacted, we hope this issue is fixed in the next AU release.
- The AC doesn't account for the reduced kill cooldown for impostors after the previous kill was prevented by the guardian angel. I can write an angry short story about this.
- There are some issues when playing with full lobbies on clients with a low MTU, as a very large (600 bytes) packet is sent upon joining. Workaround is to not have that player as a host and have this player join early and not as one of the last players. Base game issue. 2021.11.9.5 alleviated this issue by splitting up some spawn packets, but I've still heard reports of it happening.

### Closes issues

- closes #440


AeonLucid:
> The AC doesn't account for the reduced kill cooldown for impostors after the previous kill was prevented by the guardian angel. I can write an angry short story about this.

Does this mean the anticheat won't work and needs to be disabled?

YumNumm:
Ah.... Has anyone been able to confirm that this works properly?
I'm trying now but it doesn't work well...
(Error MSG:  Reliadble packer 1 (size=72) was not ack'd after 7514ms (9resends)

PR #456, opened 2022-01-21
Remove anticheat log when anticheat is disabled

simonkellly:
### Description
Does what it says on the tin


simonkellly:
"Its not a bug its a feature" -6pak

PR #469, opened 2022-03-27
Generate allcaps lobby code

rhonda:
It is causing issues when the game code contains lowercase letters, so let's make sure it is created in ALLCAP


miniduikboot:
Are you sure your fix is correct? It doesn't compile for me, because GameCode does not have a method called ToUpper and I think we already make it uppercase: https://github.com/Impostor/Impostor/blob/master/src/Impostor.Api/Games/GameCode.cs#L17

PR #474, opened 2022-04-16
Add handling for null ShipStatus

AlexejheroYTB:
### Description

This pull request fixes disconnections caused by a null ShipStatus.

#### With a null ShipStatus:
- If a player tries to enter, exit, or move between vents, they are disconnected for cheating instead of causing an exception
- When assigning tasks, the server will now assign a null ITask on each TaskInfo object instead of causing an exception
- The server will abstain from spawning players in, instead of causing an exception

I've also changed the log level for the unknown spawnable log from error to warning.

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
@@ -66,6 +66,16 @@ namespace Impostor.Server.Net.Inner.Objects.Components
                             throw new ArgumentOutOfRangeException(nameof(call), call, null);
                     }
 
+                    if (Game.GameNet.ShipStatus == null)
+                    {
+                        if (await sender.Client.ReportCheatAsync(call, "Client interacted with vent on unknown map"))
+                        {
+                            return false;
+                        }
+
+                        break;
+                    }
+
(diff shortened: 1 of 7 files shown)


miniduikboot (approved):

src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
```diff
@@ -66,6 +66,16 @@ public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, Client
...
+                    {
+                        if (await sender.Client.ReportCheatAsync(call, "Client interacted with vent on unknown map"))
+                        {
+                            return false;
+                        }
+
+                        break;
+                    }
+
                     if (!Game.GameNet.ShipStatus!.Data.Vents.TryGetValue(ventId, out var vent))
```
AeonLucid, line 79: Maybe remove `!` from `Game.GameNet.ShipStatus!.` as well.

AlexejheroYTB, line 79:
```suggestion
                    if (!Game.GameNet.ShipStatus.Data.Vents.TryGetValue(ventId, out var vent))
```

PR #491, opened 2022-08-28
Add Chat Language to game listings

miniduikboot:
This was added to the in 6.23 patch

Note that this breaks lobby listing support in 3.29 and older, but I don't suspect many players will use this version. Adding a gameVersion would complicate this code and cause an public api break


js6pak (approved):

miniduikboot:
Turns out to work in 3.29 anyway, so ignore my concern

PR #500, opened 2022-12-18
Upgrade to net7.0 and move hazel to nuget

js6pak:
Depends on https://github.com/Impostor/Impostor.Hazel/pull/1


src/Impostor.Api/GlobalUsings.cs
```diff
@@ -0,0 +1 @@
+global using Impostor.Hazel.Abstractions;
```
miniduikboot, line 1: point of this is to prevent everyone needing to change their code? does this extend to api users?

src/Impostor.Api/Net/MessageWriterExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Impostor.Api.Net;
+
+public static class MessageWriterExtensions
+{
+    public static void Serialize(this GameCode gameCode, IMessageWriter writer)
+    {
+        writer.Write(gameCode.Value);
+    }
+
+    public static void Write(this IMessageWriter writer, IInnerNetObject? innerNetObject)
```
miniduikboot, line 15: Why allow calling with a null IInnerNetObject?

miniduikboot (approved):
just nits

src/Impostor.Api/Net/MessageWriterExtensions.cs
js6pak: To mimic base game behavior

src/Impostor.Api/GlobalUsings.cs
js6pak: The point is to not generate another 3k line changes. It doesn't extend to api users as there is no good way export global usings (except nuget .props hackery)

miniduikboot: Not extending to users is absolutely fine. You're already generating the 3k removals by removing it from everything

But now that I think about it, importing a class in every single class of a program is kind of bad as well, so this change can stay

AeonLucid:
lgtm

PR #501, opened 2022-12-18
Add 2022.12.8 support

js6pak:
Todo:
- [x] GameManager netobjects
- [x] fix SyncSettings api
- [x] use filter tags in game listing


src/Impostor.Api/Innersloth/GameOptions/RoleOptions/ScientistRoleOptions.cs
```diff
@@ -0,0 +1,33 @@
+namespace Impostor.Api.Innersloth.GameOptions.RoleOptions;
+
+public class ScientistRoleOptions : IRoleOptions
+{
+    public ScientistRoleOptions(byte version)
+    {
+        Version = version;
+    }
+
+    public byte Version { get; }
```
AeonLucid, line 10: This property does not exist in the base game, and it seems unused. Why?

src/Impostor.Api/Innersloth/GameOptions/RoleOptions/ShapeshifterRoleOptions.cs
```diff
@@ -0,0 +1,37 @@
+namespace Impostor.Api.Innersloth.GameOptions.RoleOptions;
+
+public class ShapeshifterRoleOptions : IRoleOptions
+{
+    public ShapeshifterRoleOptions(byte version)
+    {
+        Version = version;
+    }
+
+    public byte Version { get; }
```
AeonLucid, line 10: This property does not exist in the base game, and it seems unused. Why?

src/Impostor.Api/Innersloth/GameOptions/RoleOptions/ScientistRoleOptions.cs
js6pak: Base game instead postfix the class with `V07`, there is no V8 yet but seems like implementing it with a property will be way cleaner

miniduikboot: We can always add this field later down the road, we don't need to anticipate what Innersloth could possibly do next

js6pak: It's part of GameOptions which is already versioned

AeonLucid: If it's already part of something that has a version, do the underlying objects really need a version too?

js6pak: Yes? How else would you handle a new field in ScientistRoleOptions for example?

AeonLucid: Basically what @miniduikboot said, but it's w/e.

AeonLucid:
Added filtering on the filter tags, but this is untested in UDP. I will also add the relevant code to Impostor.Http, which I can test.

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Impostor.Api.Net.Inner;
+using Impostor.Api.Net.Inner.Objects;
+
+namespace Impostor.Server.Net.Inner.Objects.GameManager.Logic;
+
+internal abstract class GameLogicComponent
+{
+    public virtual void OnPlayerDisconnect(IInnerPlayerControl pc)
```
js6pak, line 8: Is this needed?

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+    }
+
+    public virtual void HandleRPC(RpcCalls callId, IMessageReader reader)
```
js6pak, line 12: `HandleRpcAsync` or remove altogether until its actually needed?

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+    }
+
+    public abstract bool Serialize(IMessageWriter writer, bool initialState);
```
js6pak, line 16: Most implementations are empty anyway, make this virtual?

src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
```diff
@@ -39,6 +39,21 @@ protected T AddGameLogic<T>(T logic)
...
+        for (var i = 0; i < LogicComponents.Count; i++)
+        {
+            var component = LogicComponents[i];
+            if (component == logic)
+            {
+                return i;
+            }
+        }
+
+        return null;
```
js6pak, line 54: `return LogicComponents.IndexOf(logic)`?

src/Impostor.Server/Net/State/Game.Api.cs
```diff
@@ -20,24 +23,45 @@ public void BanIp(IPAddress ipAddress)
...
+            using var writer = MessageWriter.Get(MessageType.Reliable);
+
+            writer.StartMessage(MessageFlags.GameData);
+            Code.Serialize(writer);
+
+            writer.StartMessage(GameDataTag.DataFlag);
+            writer.WritePacked(gameManager.NetId);
+
+            writer.StartMessage((byte)gameOptionsTag);
+            GameOptionsFactory.Serialize(writer, Options);
```
js6pak, line 58: Use `LogicOptions.Serialize` here, somehow?

src/Impostor.Server/Net/State/Game.Api.cs
```diff
@@ -20,24 +23,45 @@ public void BanIp(IPAddress ipAddress)
...
         public async ValueTask SyncSettingsAsync()
         {
             if (Host?.Character == null)
             {
-                throw new ImpostorException("Attempted to set infected when the host was not spawned.");
+                throw new ImpostorException("Attempted to change settings when the host was not spawned.");
             }
 
-            using (var writer = StartRpc(Host.Character.NetId, RpcCalls.SyncSettings))
+            var gameManager = FindObjectByType<InnerGameManager>();
```
js6pak, line 33: Just cache it like every other netobject in `Game.OnSpawnAsync`?

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
AeonLucid: I just copied what the base game did. Might be okay to remove this one.

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
AeonLucid: I just copied what the base game did, maybe keep it and put a `throw NotImplementedException()` so we know when it is used in the future.

src/Impostor.Server/Net/Inner/Objects/GameManager/Logic/GameLogicComponent.cs
AeonLucid: I don't think we use any Serialize function in Impostor, should be okay to make this virtual to remove unnecessary code.

src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
```diff
@@ -0,0 +1,79 @@
...
+    private readonly ILogger<InnerGameManager> _logger;
+
+    public InnerGameManager(ICustomMessageManager<ICustomRpc> customMessageManager, Game game, ILogger<InnerGameManager> logger) : base(customMessageManager, game)
+    {
+        _logger = logger;
+
+        Components.Add(this);
+    }
+
+    protected readonly List<GameLogicComponent> LogicComponents = new();
```
js6pak, line 23: Make this private or a property?

src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
AeonLucid: Could do that, did it like this because I maybe wanted to check types and base types to fetch a tag. So instead of `GetGameLogicTag(someInstance)` you would do `GetGameLogicTag<LogicOptions>()`, that way you don't need an instance. What is your opinion on that?

src/Impostor.Server/Net/State/Game.Api.cs
AeonLucid: Right, I forgot about that.

src/Impostor.Server/Net/State/Game.Api.cs
AeonLucid: It will only replace the `GameOptionsFactory.Serialize(writer, Options);` line then. But we don't use the `LogicOptions.Serialize` call anywhere else as said by an earlier comment.

src/Impostor.Api/Innersloth/CrossplayFlags.cs
```diff
@@ -0,0 +1,7 @@
+namespace Impostor.Api.Innersloth;
+
+public enum CrossplayFlags
+{
+    Default = 4,
```
miniduikboot, line 5: This may be the default for Steam, but for Itch it is 32. I'm assuming this is following the order in Platforms.cs

src/Impostor.Api/Innersloth/GameOptions/LegacyGameOptionsData.cs
```diff
@@ -0,0 +1,263 @@
+using Impostor.Api.Innersloth.GameOptions.RoleOptions;
+
+namespace Impostor.Api.Innersloth.GameOptions;
+
+public class LegacyGameOptionsData : IGameOptions
```
miniduikboot, line 5: why do we still want this class?

src/Impostor.Api/Net/Inner/Objects/GameManager/IInnerHideAndSeekManager.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Impostor.Api.Net.Inner.Objects.GameManager.Logic.HideAndSeek;
+
+namespace Impostor.Api.Net.Inner.Objects.GameManager;
+
+public interface IInnerHideAndSeekManager
```
miniduikboot, line 5: does this interface need to inherit from IGameManager? idem IInnerNormalGameManager

src/Impostor.Api/Net/Messages/S2C/Message00HostGameS2C.cs
```diff
@@ -12,7 +12,7 @@ public static void Serialize(IMessageWriter writer, int gameCode)
             writer.EndMessage();
         }
 
-        public static GameOptionsData Deserialize(IMessageReader reader)
+        public static LegacyGameOptionsData Deserialize(IMessageReader reader)
```
miniduikboot, line 15: overzealous refactor tool?

src/Impostor.Api/Net/Inner/Objects/GameManager/IInnerHideAndSeekManager.cs
AeonLucid: Solved.

src/Impostor.Api/Net/Messages/S2C/Message00HostGameS2C.cs
js6pak: yes, but that signature is incorrect for that message in the first place

src/Impostor.Api/Innersloth/GameOptions/LegacyGameOptionsData.cs
js6pak: Impostor.Server itself doesn't really need it anymore, but is there any reason to remove it? You could use it parse old gameoptions data on disk or whatever

miniduikboot (approved):

PR #502, opened 2022-12-18
Remove redirector

js6pak:
Removes the feature no one really ever used, it probably doesn't even work
It's functionality would be better served by an external http matchmaker


AeonLucid:
I think the markdown files also require some updates to reflect this removal.

PR #505, opened 2023-01-11
Update the website for the Http Matchmaking era

miniduikboot:
### Description

This PR updates the website to return StaticHttpRegionInfo regions instead of DnsRegionInfo, as these have broken significantly in AU 2022.12.8+. Support for generating DnsRegionInfo is not retained.

Open issues:
- Not completely happy with the "generate sharable link" option, probably too niche to put it on the spotlight like this
- Documentation of the server itself is not up to date yet for StaticHttpRegionInfo, out of scope for this PR

Preview: https://mini.duikbo.at/http-regionfile/

### Closes issues

- closes the issue that StaticHttpRegionInfo files aren't well documented anywhere


script.js
```diff
@@ -97,45 +125,82 @@ function setPlatform(platform) {
...
+                    Ip: "https://matchmaker-as.among.us",
+                    Port: 443,
+                    UseDtls: true,
+                    Players: 0,
+                    ConnectionFailures: 0,
+                },
+            ],
             TranslateName: 291,
-            UseDtls: true,
         },
```
js6pak, line 187: I don't think we need these anymore, seems like they get added automatically when missing

script.js
```diff
@@ -97,45 +125,82 @@ function setPlatform(platform) {
...
             Name: name,
+            PingServer: fqdn,
+            Servers: [
+                {
+                    Name: "http-1",
+                    Ip: protocol + "://" + fqdn,
+                    Port: port,
+                    UseDtls: false, // As no custom key can be specified, we need to disable DTLS on custom servers.
+                    Players: 0,
+                    ConnectionFailures: 0,
```
js6pak, line 200: Leave these out and let them be replaced with defaults automatically?

script.js
```diff
@@ -1,4 +1,4 @@
-const DEFAULT_PORT = 22023;
+const DEFAULT_PORT = 22000;
```
js6pak, line 1: Why `22000` and not `80`?

script.js
```diff
@@ -97,45 +125,82 @@ function setPlatform(platform) {
     currentPlatform = platform;
 }
 
-function generateRegionInfo(name, ip, fqdn, port) {
+function setPortIfDefault(protocol) {
+    const oldPort = protocol == "http" ? "443" : "22000";
+    const newPort = protocol == "http" ? "22000" : "443";
```
js6pak, line 130: Use the constant `DEFAULT_PORT` here? (and split it to `DEFAULT_HTTP_PORT` and `DEFAULT_HTTPS_PORT` I guess)

js6pak (changes requested):
Also, you've used `==` instead `===` everywhere, so we should probably setup eslint to ban that

script.js
miniduikboot: 22000 is the default port that Impostor.Http uses and therefore the port that people who don't use a reverse proxy will use.

script.js
miniduikboot: yeah, that'd be cleaner, I'll change it

script.js
miniduikboot: They do get automatically added, but the lack of official servers can confuse some people who are not aware of that happening.

I shared a template json file in the discord a few days ago and I already received a few comments from people that were missing the original servers. Adding the bytes doesn't hurt and as it reduces confusion I think we should leave them in

script.js
miniduikboot: do the servers get filled in automaticallly?

script.js
js6pak: I mean more the maintaining burden if among us ever changes the addresses or anything.
As for confusion we could add a comment in json if you really want? Personally I don't think it's really a problem.

script.js
js6pak: Servers? I mean `Players` and `ConnectionFailures`, they both default to `0` anyway. Either way doesn't really matter, just a nitpick (and me wanting to keep this js script to the bare minimum).

script.js
miniduikboot: I don't expect regular changes to the addresses, and even if upstream changes these addresses, old region files will be silently updated anyway. New addresses will likely also require changes in our region file generator

js6pak (approved):

miniduikboot:
I'm concerned about the lack of default servers and if that will confuse people, but let's just merge it and see

PR #508, opened 2023-01-19
systemd template and instructions

bkloosterboer:
### Description

Adds a template and instructions for starting and enabling Impostor using systemd.

### DIFF
--- a/docs/Server-configuration.md
+++ b/docs/Server-configuration.md
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
 IMPOSTOR_AntiCheat__Enabled=true
 IMPOSTOR_AntiCheat__BanIpFromGame=true
 ```
+
+# systemd
+
+Linux users will likely want to create a systemd service definition to manage running the Impostor application.  Benefits to using a systemd service include the following:
+
+- The server process can automatically restart if the process dies
+- Impostor can be set to start on boot (handy if you do system updates regularly)
+- You can use systemd targets to ensure the process doesn't start until the system dependencies are met
+
+Using your favorite text editor, create the following service file:
+
+```
+sudo vim /etc/systemd/system/impostor.service
+```
+
+Modify the following template to your needs and populate the service file.
+
+```
+[Unit]
+Description=Impostor private Among Us server - https://github.com/Impostor/Impostor
+
+[Service]
+# Directory where Impostor is installed
+WorkingDirectory=/opt/impostor
+# Path to Impostor binary
+ExecStart=/opt/impostor/Impostor.Server
+Restart=always
+# Restart service after 10 seconds if the dotnet service crashes
+RestartSec=10
+KillSignal=SIGINT
+SyslogIdentifier=log.impostor
+
+# User and group that will run the Impostor server process -- ensure that user has execute on the binary
+User=impostor
+Group=impostor
+TimeoutStopSec=10
+
+[Install]
+# Wait until most system services have started before starting Impostor
+WantedBy=multi-user.target
+```
+
+Reload all systemd unit files to source in the new file you've created.
+
+```
+sudo systemctl daemon-reload
+```
+
+Start the service and verify that it starts correctly.  Hint: Use the Q key to quit out of journalctl.
+
+```
+sudo systemctl start impostor.service
+```
+
+```
+sudo journalctl -u impostor.service
+```
+
+If everything has started correctly, ensure the service is set to start on boot.
+
+```
+sudo systemctl enable impostor.service
+```
--- a/docs/Server-configuration.md
+++ b/docs/Server-configuration.md
@@ -105,7 +105,7 @@ Linux users will likely want to create a systemd service definition to manage ru
 Using your favorite text editor, create the following service file:
 
 ```
-sudo vim /etc/systemd/system/impostor.service
+sudoedit /etc/systemd/system/impostor.service
 ```
 
 Modify the following template to your needs and populate the service file.
@@ -122,10 +122,10 @@ ExecStart=/opt/impostor/Impostor.Server
 Restart=always
 # Restart service after 10 seconds if the dotnet service crashes
 RestartSec=10
-KillSignal=SIGINT
+KillSignal=SIGTERM
 SyslogIdentifier=log.impostor
 
-# User and group that will run the Impostor server process -- ensure that user has execute on the binary
+# User and group that will run the Impostor server process -- ensure that user is allowed to execute the binary
 User=impostor
 Group=impostor
 TimeoutStopSec=10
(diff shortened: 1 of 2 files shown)


docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+Linux users will likely want to create a systemd service definition to manage running the Impostor application.  Benefits to using a systemd service include the following:
+
+- The server process can automatically restart if the process dies
+- Impostor can be set to start on boot (handy if you do system updates regularly)
+- You can use systemd targets to ensure the process doesn't start until the system dependencies are met
+
+Using your favorite text editor, create the following service file:
+
+```
+sudo vim /etc/systemd/system/impostor.service
```
miniduikboot, line 108: ooh, getting people stuck in vim, a true classic.

Maybe it's nicer to use `sudoedit` instead? Then it automatically opens in whatever editor they have configured

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+```
+
+Modify the following template to your needs and populate the service file.
+
+```
+[Unit]
+Description=Impostor private Among Us server - https://github.com/Impostor/Impostor
```
miniduikboot, line 115: I'm looking at my own unit file (probably should've upstreamed that sooner but oops) and I also have two exta lines here:

```
After=syslog.target
After=network.target
```

This seems to be orthagonal to WantedBy, should I add these lines here or are they sort of obsolete?

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+WorkingDirectory=/opt/impostor
+# Path to Impostor binary
+ExecStart=/opt/impostor/Impostor.Server
+Restart=always
+# Restart service after 10 seconds if the dotnet service crashes
+RestartSec=10
+KillSignal=SIGINT
+SyslogIdentifier=log.impostor
+
+# User and group that will run the Impostor server process -- ensure that user has execute on the binary
```
miniduikboot, line 128:
```suggestion
# User and group that will run the Impostor server process -- ensure that user is allowed to execute the binary
```

I think this is more clear to the more casual linux user

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+[Service]
+# Directory where Impostor is installed
+WorkingDirectory=/opt/impostor
+# Path to Impostor binary
+ExecStart=/opt/impostor/Impostor.Server
+Restart=always
+# Restart service after 10 seconds if the dotnet service crashes
+RestartSec=10
+KillSignal=SIGINT
+SyslogIdentifier=log.impostor
```
miniduikboot, line 126: Hmm, what is the effect of this line? Can you share an example log line?

For example, this is a log entry on my server:
```
Jan 19 23:15:24 euhtzfsn1 Impostor.Server[11330]: [23:15:24 WRN] Received RpcFlag for unregistered NetId 4.
```

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+
+[Service]
+# Directory where Impostor is installed
+WorkingDirectory=/opt/impostor
+# Path to Impostor binary
+ExecStart=/opt/impostor/Impostor.Server
+Restart=always
+# Restart service after 10 seconds if the dotnet service crashes
+RestartSec=10
+KillSignal=SIGINT
```
miniduikboot, line 125: Hmm, I don't have this value specified in my config file. The default KillSignal seems to be SIGTERM, does SIGINT make a difference here? I tested it on my pc and in both cases the client was disconnected gracefully

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+WantedBy=multi-user.target
+```
+
+Reload all systemd unit files to source in the new file you've created.
+
+```
+sudo systemctl daemon-reload
+```
+
+Start the service and verify that it starts correctly.  Hint: Use the Q key to quit out of journalctl.
```
miniduikboot, line 144: Hmm, the hint only applies to the second command. Suggestion: add a small explanation after the first command and move the hint to there

docs/Server-configuration.md
```diff
@@ -93,3 +93,66 @@ IMPOSTOR_Server__ListenPort=22023
...
+- Impostor can be set to start on boot (handy if you do system updates regularly)
+- You can use systemd targets to ensure the process doesn't start until the system dependencies are met
+
+Using your favorite text editor, create the following service file:
+
+```
+sudo vim /etc/systemd/system/impostor.service
+```
+
+Modify the following template to your needs and populate the service file.
```
miniduikboot, line 111: Should we put the template in here or should we put it as a separate file that people can just copypaste? (or download from github and use directly)

miniduikboot (approved):
Thanks for your suggestion. I probably should have added a section like this much earlier.

I've left some remarks, let me know if you agree and I can apply them for you if you want. I'm in favor of merging this PR, but I'd like to merge it as a perfect 10 and not an 6

docs/Server-configuration.md
bkloosterboer: that sounds reasonable to me

docs/Server-configuration.md
bkloosterboer: WantedBy=multiuser.target adds the multiuser target as a dependency.  I think it's a slightly cleaner way of doing it.  Ultimately your way would also likely work.

docs/Server-configuration.md
bkloosterboer: That sounds good to me

docs/Server-configuration.md
bkloosterboer: > Jan 20 02:25:39 amongus log.impostor[6741]: [02:25:39 INF] Client 3 disconnecting, reason: The remote sent a disconnect request

docs/Server-configuration.md
bkloosterboer: Let's switch it to SIGTERM.  SIGINT sends control + c and after doing some reading, SIGTERM makes more sense.

docs/Server-configuration.md
bkloosterboer: will do

docs/Server-configuration.md
miniduikboot: It's not `After` or `WantedBy`, you want both:
- After specifies this unit must start when a few other units have also started
- WantedBy gives a deadline: this unit must start for the system to consider itself to be fully started.

docs/Server-configuration.md
miniduikboot: ah okay, it just changes the name then. Not sure why we'd need the `log.` prefix then, maybe remove that prefix?

miniduikboot:
Thanks for your PR, I hope this helps a few people with setting up a systemd unit for their servers!

PR #509, opened 2023-01-19
Add Impostor.Http to the Docker image

miniduikboot:
Adding plugins to the docker image is pretty hard as you need to install ASP.Net Core and add the plugin. As Impostor.Http has become pretty much mandatory, install the plugin by default.

Depends on https://github.com/Impostor/Impostor/pull/507

### Closes issues

- helps various people in the Discord with Docker issues


js6pak (approved):
I hate it, but I guess its fine until we merge Impostor.Http to base

miniduikboot:
I'm also not a big fan, but this is the pragmatic fix that we can do today.

PR #518, opened 2023-03-06
Add conditions to the GameFull judgment

Huier-Huang:
### Description

Add conditions to the GameFull judgment to avoid showing that the game room is full but has actually started


Huier-Huang:
I think it is necessary to add that, in my subjective opinion, GameFull should only be returned when the game has not yet started

PR #531, opened 2023-07-14
Add Version 2023.7.11 and Support TownOfHost

Huier-Huang:


miniduikboot:
Hey there, sorry for the lack of replies, I had a busy week.

I've not added the 2222.0.0.0 version to my PR because of two issues that make it not usable in Impostor's base configuration:
1. Impostor's anticheat does not support TOH, so it needs to be disabled
2. Impostor by default does not allow crossplay between different versions, which makes it impossible for a TOH player to host for non-TOH users
Using the compatibility options in Impostor it is possible to fix these issues right now. I'm hoping to fix these issues in Impostor itself soon-ish but can't give an ETA for it

PR #532, opened 2023-07-15
Add ModConfig And Support TownOfHost

Huier-Huang:
### Closes issues

- closes #


miniduikboot:
Hi there,

Thank you for your contribution. As mentioned in #531 we planned to implement a more generic version that can be more easily extended in the future and is not limited to this specific network version. It'll also make Impostor closer to the behaviour of the official servers. Unfortunately this took a few days to develop, so you were able to submit this solution before us. 
I think that our solution is more flexible and can also be used to make other vanilla game versions compatible with each other, which is a problem we had in the past and will continue to have as long as Innersloth releases minor patches that are compatible with each other network-wise, but have a different network version. Your PR assumes that there is a specific game version for mods, but this may become irrelevant with the next version if Innersloth decides on a different solution to disable server authority on the official servers. Sorry for wasting your effort.

PR #533, opened 2023-07-16
Merge Impostor.Http into core

js6pak:


src/Impostor.Api/Config/HttpServerConfig.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// <summary>
+    /// Gets or sets the port the HTTP Matchmaking server will listen on.
+    /// </summary>
+    /// For port forwarding purposes, this is a TCP port.
+    public ushort ListenPort { get; set; } = 22023;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether Https is enabled or not.
+    /// </summary>
+    /// If true, also set CertificatePath.
```
miniduikboot, line 31:
```suggestion
    /// If true, CertificatePath needs to be set to a SSL certificate.
```

src/Impostor.Server/Http/GamesController.cs
```diff
@@ -0,0 +1,154 @@
...
+    [HttpPut]
+    public IActionResult Put()
+    {
+        return Ok(_hostServer);
+    }
+
+    private static uint ConvertAddressToNumber(IPAddress address)
+    {
+#pragma warning disable CS0618 // Among Us only supports IPv4
+        return (uint)address.Address;
```
miniduikboot, line 85: .Address [throws if provided an IPv6 address](https://learn.microsoft.com/en-us/dotnet/api/system.net.ipaddress.address?view=net-7.0#exceptions), maybe check if it is an IPv4 address using AddressFamily and if not, log something helpful instead of throwing an exception

src/Impostor.Server/Http/ListingManager.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// Find listings that match the requested settings.
+    /// </summary>
+    /// <param name="ctx">The context of this http request.</param>
+    /// <param name="map">The selected maps.</param>
+    /// <param name="impostorCount">The amount of impostors. 0 is any.</param>
+    /// <param name="language">Chat language of the game.</param>
+    /// <param name="maxListings">Maximum amount of games to return.</param>
+    /// <returns>Listings that match the required criteria.</returns>
+    public IEnumerable<IGame> FindListings(HttpContext ctx, int map, int impostorCount, GameKeywords language, int maxListings = 10)
```
miniduikboot, line 39: This method is a dupe of GameManager.FindListings (although we improved it afterwards)
Do we want to merge these functions back together again or should we remove game list requesting over udp?

Merging may be difficult because of the ListingFilter getting an HttpContext as an argument, which was the whole reason for the split

src/Impostor.Server/Http/TokenController.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    /// <summary>
+    /// Get an authentication token.
+    /// </summary>
+    /// <param name="request">Token parameters that need to be put into the token.</param>
+    /// <returns>A bare minimum authentication token that the client will accept.</returns>
+    [HttpPost]
+    public IActionResult GetToken([FromBody] TokenRequest request)
+    {
+        var token = new Token(new TokenPayload(request.ProductUserId, request.ClientVersion), "impostor_http_was_here");
```
miniduikboot, line 24:
```suggestion
        var token = new Token(new TokenPayload(request.ProductUserId, request.ClientVersion), "impostor_was_here");
```

the plugin is dead

src/Impostor.Server/Http/GamesController.cs
js6pak: It would throw in https://github.com/Impostor/Impostor/blob/7b341bea147b763ef879e13e82c072dcf5ae6e25/src/Impostor.Api/Utils/IpUtils.cs#L32-L36 before it gets to this point

src/Impostor.Server/Http/ListingManager.cs
js6pak: I was going to ask about this but forgot, imo we should just remove udp matchmaking to make the code cleaner

miniduikboot: There is not actually much we can remove, joining via code or creating a new game still happens through udp mm. Only the gamelist stuff can actually be removed

src/Impostor.Server/Net/Client.cs
```diff
@@ -260,8 +259,8 @@ public override async ValueTask HandleMessageAsync(IMessageReader reader, Messag
 
                 case MessageFlags.GetGameListV2:
                 {
-                    Message16GetGameListC2S.Deserialize(reader, out var options, out _, out _, out var filterOptions);
-                    await OnRequestGameListAsync(options, filterOptions);
+                    Message16GetGameListC2S.Deserialize(reader, out _, out _, out _, out _);
```
miniduikboot, line 262: why even deserialize if you're going to throw the message in the trash and disconnect them anyway?

miniduikboot (commented):
I think this will need some doc changes too, but I'll do that in a separate PR

src/Impostor.Api/Config/HttpServerConfig.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// Gets or sets the port the HTTP Matchmaking server will listen on.
+    /// </summary>
+    /// For port forwarding purposes, this is a TCP port.
+    public ushort ListenPort { get; set; } = 22023;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether Https is enabled or not.
+    /// </summary>
+    /// If true, CertificatePath needs to be set to a SSL certificate.
+    public bool UseHttps { get; set; }
```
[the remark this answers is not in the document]
js6pak: Yeah, I don't really see a use case where terminating SSL in a reverse proxy isn't both easier and better.

src/Impostor.Server/Plugins/PluginLoader.cs
miniduikboot, line 1: Line 30-41: Do we still need to add the ASP.NET Core path or is that now automatically included?

js6pak: Actually ASP.NET Core runtime is now required so that code is no longer needed.

miniduikboot (approved):

miniduikboot (approved):

PR #534, opened 2023-07-18
Improve cross-compatibility

miniduikboot:
### Description

Official servers are allowing multiple game versions to crossplay. Impostor currently has a binary toggle that allows or forbids *all* crossplay. This PR adds the so-called CompatibiltyManager which puts all supported network versions in groups: if two versions are in the same group, they are allowed to play together.

This feature is also exposed as API so that plugins can mess this up on runtime without requiring server restarts every time Innersloth releases a small patch which bumps the network version but does not change the network protocol enough.

Also bump the minor version because we're introducing API and changing behaviour compared to previous versions.

### Closes issues

- closes #532

### DIFF
--- a/src/Impostor.Api/Innersloth/GameVersion.cs
+++ b/src/Impostor.Api/Innersloth/GameVersion.cs
@@ -5,6 +5,10 @@ namespace Impostor.Api.Innersloth
 {
     public readonly struct GameVersion : IEquatable<GameVersion>, IComparable<GameVersion>, IComparisonOperators<GameVersion, GameVersion, bool>
     {
+        private const int YearMask = 25000;
+        private const int MonthMask = 1800;
+        private const int DayMask = 50;
+
         public GameVersion(int value)
         {
             Value = value;
@@ -12,10 +16,18 @@ namespace Impostor.Api.Innersloth
 
         public GameVersion(int year, int month, int day, int revision = 0)
         {
-            Value = (year * 25000) + (month * 1800) + (day * 50) + revision;
+            Value = (year * YearMask) + (month * MonthMask) + (day * DayMask) + revision;
         }
 
         public int Value { get; }
+
+        public int Year => Value / YearMask;
+
+        public int Month => (Value % YearMask) / MonthMask;
+
+        public int Day => ((Value % YearMask) % MonthMask) / DayMask;
+
+        public int Revision => Value % DayMask;
 
         public static bool operator ==(GameVersion left, GameVersion right) => left.Value == right.Value;
 
@@ -32,19 +44,19 @@ namespace Impostor.Api.Innersloth
         public void GetComponents(out int year, out int month, out int day, out int revision)
         {
             var value = Value;
-            year = value / 25000;
-            value %= 25000;
-            month = value / 1800;
-            value %= 1800;
-            day = value / 50;
-            revision = value % 50;
+            year = value / YearMask;
+            value %= YearMask;
+            month = value / MonthMask;
+            value %= MonthMask;
+            day = value / DayMask;
+            revision = value % DayMask;
         }
 
         public GameVersion ExtractDisableServerAuthority(out bool disableServerAuthority)
         {
             const int DisableServerAuthorityFlag = 25;
 
-            var revision = Value % 50;
+            var revision = Value % DayMask;
             if (revision >= DisableServerAuthorityFlag)
             {
                 disableServerAuthority = true;
--- a/src/Impostor.Api/Innersloth/GameVersion.cs
+++ b/src/Impostor.Api/Innersloth/GameVersion.cs
@@ -1,20 +1,84 @@
+using System;
+using System.Numerics;
+
 namespace Impostor.Api.Innersloth
 {
-    public static class GameVersion
+    public readonly struct GameVersion : IEquatable<GameVersion>, IComparable<GameVersion>, IComparisonOperators<GameVersion, GameVersion, bool>
     {
-        public static int GetVersion(int year, int month, int day, int revision = 0)
+        public GameVersion(int value)
         {
-            return (year * 25000) + (month * 1800) + (day * 50) + revision;
+            Value = value;
         }
 
-        public static void ParseVersion(int version, out int year, out int month, out int day, out int revision)
+        public GameVersion(int year, int month, int day, int revision = 0)
         {
-            year = version / 25000;
-            version %= 25000;
-            month = version / 1800;
-            version %= 1800;
-            day = version / 50;
-            revision = version % 50;
+            Value = (year * 25000) + (month * 1800) + (day * 50) + revision;
+        }
+
+        public int Value { get; }
+
+        public static bool operator ==(GameVersion left, GameVersion right) => left.Value == right.Value;
+
+        public static bool operator !=(GameVersion left, GameVersion right) => left.Value != right.Value;
+
+        public static bool operator >(GameVersion left, GameVersion right) => left.Value > right.Value;
+
+        public static bool operator >=(GameVersion left, GameVersion right) => left.Value >= right.Value;
+
+        public static bool operator <(GameVersion left, GameVersion right) => left.Value < right.Value;
+
+        public static bool operator <=(GameVersion left, GameVersion right) => left.Value <= right.Value;
+
+        public void GetComponents(out int year, out int month, out int day, out int revision)
+        {
+            var value = Value;
+            year = value / 25000;
+            value %= 25000;
+            month = value / 1800;
+            value %= 1800;
+            day = value / 50;
+            revision = value % 50;
+        }
+
+        public GameVersion ExtractDisableServerAuthority(out bool disableServerAuthority)
+        {
+            const int DisableServerAuthorityFlag = 25;
+
+            var revision = Value % 50;
+            if (revision >= DisableServerAuthorityFlag)
+            {
+                disableServerAuthority = true;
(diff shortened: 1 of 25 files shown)


src/Impostor.Server/Net/Manager/CompatibilityManager.cs
```diff
@@ -0,0 +1,100 @@
...
+            new(GameVersion.GetVersion(2023, 1, 11), GameVersion.GetVersion(2023, 1, 11), true), // 2023.3.28s
+            new(GameVersion.GetVersion(2023, 3, 13), GameVersion.GetVersion(2023, 1, 11), true), // 2023.3.28a
+            new(GameVersion.GetVersion(2023, 4, 21), GameVersion.GetVersion(2023, 1, 11), true), // 2023.6.13
+
+            new(GameVersion.GetVersion(2023, 5, 20), GameVersion.GetVersion(2023, 5, 20), true), // 2023.7.11
+            new(GameVersion.GetVersion(2222, 0, 0), GameVersion.GetVersion(2023, 5, 20), false), // 2023.7.11 for host-only mods
+        };
+
+        // Map from client version to compatibility group
+        private readonly Dictionary<int, int> _supportMap = new Dictionary<int, int>();
```
js6pak, line 27:
```suggestion
        private readonly Dictionary<int, int> _supportMap = new();
```

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
```diff
@@ -0,0 +1,100 @@
...
+        private readonly ILogger<CompatibilityManager> _logger;
+        private readonly bool initializationFinished;
```
js6pak, line 29:
```suggestion
        private readonly bool _initializationFinished;
```

But also do we even need this? Wouldn't it be better to split `AddSupportedVersion` into a private and public versions?

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
```diff
@@ -0,0 +1,100 @@
...
+            foreach (var compatData in DefaultSupportedVersions)
+            {
+                AddSupportedVersion(compatData.GameVersion, compatData.CompatGroup, compatData.IncludeInSupportRange);
+            }
+
+            ModifiedByUser = false;
+            initializationFinished = true;
+        }
+
+        internal bool ModifiedByUser { get; private set; }
```
js6pak, line 46: Can't this be private?

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+        }
+
+        public bool RemoveSupportedVersion(int removedVersion)
+        {
+            return _supportMap.Remove(removedVersion);
+        }
+
+        internal record CompatData(int GameVersion, int CompatGroup, bool IncludeInSupportRange);
```
js6pak, line 98: Shouldn't `IncludeInSupportRange` default to true?

src/Impostor.Server/Net/State/Game.Incoming.cs
```diff
@@ -156,11 +156,13 @@ private async ValueTask<GameJoinResult> AddClientSafeAsync(ClientBase client)
             if (_compatibilityConfig.AllowVersionMixing == false &&
                 this.Host != null && client.GameVersion != this.Host.Client.GameVersion)
             {
-                if (client.GameVersion < this.Host.Client.GameVersion)
+                _compatibilityManager.TryGetCompatibilityGroup(client.GameVersion, out var clientCompatGroup);
```
js6pak, line 159: I know all clients should have a compat group at this point but it still feels wrong to just ignore the return value

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            new(GameVersion.GetVersion(2022, 12, 2), GameVersion.GetVersion(2022, 12, 2), true), // 2023.2.28
+
+            new(GameVersion.GetVersion(2023, 1, 11), GameVersion.GetVersion(2023, 1, 11), true), // 2023.3.28s
+            new(GameVersion.GetVersion(2023, 3, 13), GameVersion.GetVersion(2023, 1, 11), true), // 2023.3.28a
+            new(GameVersion.GetVersion(2023, 4, 21), GameVersion.GetVersion(2023, 1, 11), true), // 2023.6.13
+
+            new(GameVersion.GetVersion(2023, 5, 20), GameVersion.GetVersion(2023, 5, 20), true), // 2023.7.11
+            new(GameVersion.GetVersion(2222, 0, 0), GameVersion.GetVersion(2023, 5, 20), false), // 2023.7.11 for host-only mods
+        };
```
js6pak, line 24: There's a lot of repetitiveness here. Maybe make the compat group an enum instead? Or store the game versions in static fields and reference them here

src/Impostor.Api/Net/Manager/ICompatibilityManager.cs
```diff
@@ -0,0 +1,61 @@
...
+         * Get the compatibility group of a certain client version. If two versions are in the same compatibility group,
+         * they can play a game together without issues.
+         * </summary>
+         * <param name="clientVersion">The client version to check for.</param>
+         * <param name="compatGroup">If the result is Compatible, the compat group.</param>
+         * <returns>
+         * Whether this version is supported by the server at the moment and if not, whether it is too old or too new.
+         * </returns>
+         */
+        public VersionCompareResult TryGetCompatibilityGroup(int clientVersion, out int? compatGroup);
```
js6pak, line 30: We only ever use this to either get the compare result or the compat group, shouldn't this be split into 2 methods - Compare and GetCompatiblityGroup? Eventually only use this inside `ClientManager` and pass the compat group down to `Game` inside `IClient`

js6pak (changes requested):
Also, please use `///`  instead `/**`

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
miniduikboot: If I'd make it an enum, removing a compatibility groups would be a breaking api change. Maybe it'd be better to make it an GameVersion array array where each element of the GameVersion array would be members of the same compatibility group?

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
miniduikboot: We don't need it, but considering this API is a massive footgun if used incorrectly, I wanted to print a warning if it is. 

Splitting into private/public would be a dupe and considering `private` in C# to me seems not much more than a suggestion, I don't think this would be useful

src/Impostor.Server/Net/State/Game.Incoming.cs
miniduikboot: Yeah, technically you could race this using the API, so I should check here again if I'm not comparing against null

src/Impostor.Api/Net/Manager/ICompatibilityManager.cs
miniduikboot: Splitted this into two methods:
- CanConnectToServer compares one client to what the server supports
- CanJoinGame compares two clients to each other to see if they can join a game together.
Original method is poof unless someone has a good usecase for it

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
miniduikboot: removed it altogether

src/Impostor.Server/Net/Manager/CompatibilityManager.cs
miniduikboot: Sticking to this for now, yes it's a bit boilerplate but pretty reviewable

src/Impostor.Api/Net/Manager/ICompatibilityManager.cs
```diff
@@ -18,37 +19,50 @@ public enum VersionCompareResult
...
 
         /// <summary>Check if a player can join an existing game.</summary>
         /// <param name="hostVersion">The client version of the host.</param>
-        /// <param name="playerVersion">The client version of the player that is joining.</param>
+        /// <param name="clientVersion">The client version of the player that is joining.</param>
         /// <returns>
         /// - GameJoinError.None if everything is OK
         /// - GameJoinError.ClientOutdated if the player runs a too old game version
-        /// - GameJoinError.ClientTooNew if the player runs a too new game version
+        /// - GameJoinError.ClientTooNew if the player runs a too new game version.
```
miniduikboot, line 42: consistency: the other items also don't end on a `.`, remove this one or introduce it to the rest too

src/Impostor.Api/Innersloth/GameVersion.cs
```diff
@@ -1,20 +1,84 @@
...
+            var value = Value;
+            year = value / 25000;
+            value %= 25000;
+            month = value / 1800;
+            value %= 1800;
+            day = value / 50;
+            revision = value % 50;
+        }
+
+        public GameVersion ExtractDisableServerAuthority(out bool disableServerAuthority)
```
miniduikboot, line 43: this method is doing two things, it may be cleaner if you split this up into two items:
- normalizing the version to remove the dsa flag
- getting whether a specific version has the dsa flag set

miniduikboot:
LGTM, feel free to merge

PR #543, opened 2023-09-06
Create Impostor.Api-Events.md

maxwo007:
Try to write a documentation about the API events.

### Closes issues

- closes #


miniduikboot (changes requested):
Hi there, thanks for your PR. Are you familiar with the XML documentation comments system of C#? (See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/ for an explanation). Putting comments in this format allows IDE's to autocomplete it, while we can also export them to a website using something like docfx. I think that placing XML documentation comments in these classes removes the need for a file such as this one.

We already have *some* documentation comments in some places, but I was a bit shocked at the amount of XML documentation comments in there when I looked through it yesterday. So improving the state of the documentation comments is something that could be useful.

What do you think about this?

maxwo007:
Thanks for your feedback. I tried to implement the XML doc system and opened a new pull request.

PR #548, opened 2023-10-30
Source generate map data and enums

js6pak:
Source generates map data classes and large enums from json dumped with [Dumpostor](https://github.com/Impostor/Dumpostor).

One open question is whether we want to add a compatibility transformation for enums like `RpcCalls` for backwards compatibility or use the new map as an occasion to break and start being in sync with base game names.

<details>
  <summary>Example generated map data (<code>SkeldData.cs</code>)</summary>

```csharp
// <auto-generated>
#nullable enable

using System.Collections.Generic;
using System.Numerics;

namespace Impostor.Api.Innersloth.Maps;

public sealed class SkeldData : MapData
{
    internal SkeldData()
    {
        Vents = new Dictionary<int, VentData>
        {
            [0] = new(this, 0, "AdminVent", new Vector2(2.544f, -9.955201f), left: 2, right: 1),
            [1] = new(this, 1, "BigYVent", new Vector2(9.384f, -6.438f), left: 0, right: 2),
            [2] = new(this, 2, "CafeVent", new Vector2(4.2588f, -0.27600002f), left: 0, right: 1),
            [3] = new(this, 3, "ElecVent", new Vector2(-9.7764f, -8.034f), left: 5, right: 6),
            [4] = new(this, 4, "LEngineVent", new Vector2(-15.288f, 2.52f), left: 11),
            [5] = new(this, 5, "SecurityVent", new Vector2(-12.534f, -6.9492f), left: 6, right: 3),
            [6] = new(this, 6, "MedVent", new Vector2(-10.608001f, -4.176f), left: 3, right: 5),
            [7] = new(this, 7, "WeaponsVent", new Vector2(8.820001f, 3.3240001f), right: 12),
            [8] = new(this, 8, "ReactorVent", new Vector2(-20.796001f, -6.9528003f), right: 9),
            [9] = new(this, 9, "REngineVent", new Vector2(-15.2508f, -13.656001f), left: 8),
            [10] = new(this, 10, "ShieldsVent", new Vector2(9.5232f, -14.337601f), right: 13),
            [11] = new(this, 11, "UpperReactorVent", new Vector2(-21.876f, -3.0516002f), right: 4),
            [12] = new(this, 12, "NavVentNorth", new Vector2(16.008001f, -3.1680002f), left: 7),
            [13] = new(this, 13, "NavVentSouth", new Vector2(16.008001f, -6.3840003f), left: 10),
        }.AsReadOnly();

        Tasks = new Dictionary<int, TaskData>
        {
[...]

PR #549, opened 2023-11-06
Feature/2023.10.24

miniduikboot:
### Description

Update Impostor to version 2023.10.24 of Among Us.

Excluded are changes to autogen map data (see #548 )


src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -688,15 +749,50 @@ private async ValueTask<bool> HandleMurderPlayer(ClientPlayer sender, IInnerPlay
 
             if (target != null && !target.PlayerInfo.IsDead)
             {
-                ((InnerPlayerControl)target).Die(DeathReason.Kill);
-                await _eventManager.CallAsync(new PlayerMurderEvent(Game, sender, this, target));
+                if ((result & (MurderResultFlags.FailedError | MurderResultFlags.FailedProtected)) == 0)
```
js6pak, line 752: 1. Wouldn't it be easier to check for `Succeeded`?
2. We do this check in 2 places, so we probably should make a `MurderResultFlags#IsFailed` extension method

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -420,6 +428,30 @@ internal void Die(DeathReason reason)
...
+                var guardianAngelOpts = (GuardianAngelRoleOptions)opts.RoleOptions.Roles[RoleTypes.GuardianAngel].RoleOptions;
+                var duration = guardianAngelOpts.ProtectionDurationSeconds;
+                var protectionExpiresAt = ProtectedOn.Value.AddSeconds(duration);
+                return protectionExpiresAt >= _dateTimeProvider.UtcNow;
+            }
+            else
+            {
+                return false;
+            }
+        }
```
js6pak, line 453: 1. 🍝
2. Parameter-less method starting with `Is` sounds wrong, shouldn't this be a property?

js6pak (changes requested):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -688,15 +749,50 @@ private async ValueTask<bool> HandleMurderPlayer(ClientPlayer sender, IInnerPlay
...
+
+                await _eventManager.CallAsync(new PlayerMurderEvent(Game, sender, this, target, result));
             }
 
             IsMurdering = null;
 
             return true;
         }
 
+        private async ValueTask<bool> HandleCheckProtect(ClientPlayer sender, IInnerPlayerControl? target)
```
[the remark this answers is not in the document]
miniduikboot: it's  just an internal method anyway

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: 1. No, because it also succeeded if it was decided by the host
2. Not opposed. Handling this bitflag is annoying

js6pak: 1. I mean `(result & MurderResultFlags.Succeeded) != 0`

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: 1. I welcome improvements to this method. It's verbose due to how the options api works
2. We could turn this into a field called `Protected` but eh

js6pak (approved):

PR #550, opened 2023-11-16
Experimental: Add lock on game objects

miniduikboot:
### Description

We see rare problems where multiple messages on the same game arriving at the same game cause issues. This commit adds a coarse-grained lock on all interactions with a game by adding a binary semaphore.

Because manually adding try{} finally{} blocks seems verbose, I added a RAII-style LockGuard that locks the semaphore on creation and drops it when disposed.

### Closes issues

- closes #


js6pak (approved):

miniduikboot:
I recall this PR didn't solve the problems it was designed to fix, do we still want to push this forward?

miniduikboot:
I'm running into issues with RoleOptionsCollection throwing ConcurrentModificationErrors, so I'm going to revive this PR

PR #559, opened 2024-01-24
Make Impostor's AC modular

miniduikboot:
### Description

Currently Impostor's anticheat is one giant monolithic block: server owners have to pick between either enabling or disabling all checks. Because there are a few checks that instantly break when client mods are involved (like unknown RPC calls), disabling the anticheat outright is a popular first step after setup. By adding a categorization like this it should be possible to add a level of anticheat that should work for clientside and host-only mods as well.

### Open issues with this PR:

- [x] Discuss granularity of the current options. Do we have too much or not enough options? Do the categorizations make sense? => I think this is fine, we can always iterate on this some more
- [x] Write documentation on the new options
- [x] Consider adding more checks => outside of this PR's scope


src/Impostor.Server/Net/Client.cs
```diff
@@ -35,14 +35,72 @@ public Client(ILogger<Client> logger, IOptions<AntiCheatConfig> antiCheatOptions
...
-        public override async ValueTask<bool> ReportCheatAsync(CheatContext context, string message)
+        public override async ValueTask<bool> ReportCheatAsync(CheatContext context, CheatCategory category, string message)
         {
             if (!_antiCheatConfig.Enabled)
             {
                 return false;
             }
 
-            _logger.LogWarning("Client {Name} ({Id}) was caught cheating: [{Context}] {Message}", Name, Id, context.Name, message);
+            if (_antiCheatConfig.ForbidHostCheating && Player != null && Player.IsHost)
```
js6pak, line 45: Shouldn't `ForbidHostCheating` be negated here?
Also I think `ExemptHost` as a name made more sense and it's fine to make it different than others because it's the only one that isn't related to categories.

src/Impostor.Server/Net/Client.cs
```diff
@@ -35,14 +35,72 @@ public Client(ILogger<Client> logger, IOptions<AntiCheatConfig> antiCheatOptions
...
+
+                    break;
+                case CheatCategory.Target:
+                    if (!_antiCheatConfig.EnableTargetChecks)
+                    {
+                        return false;
+                    }
+
+                    break;
+            }
```
js6pak, line 101:
```suggestion
            var isCategoryEnabled = category switch
            {
                CheatCategory.ProtocolExtension => _antiCheatConfig.ForbidProtocolExtensions,
                CheatCategory.GameFlow => _antiCheatConfig.EnableGameFlowChecks,
                CheatCategory.MustBeHost => _antiCheatConfig.EnableMustBeHostChecks,
                CheatCategory.Limit => _antiCheatConfig.EnableLimitChecks,
                CheatCategory.Ownership => _antiCheatConfig.EnableOwnershipChecks,
                CheatCategory.Role => _antiCheatConfig.EnableRoleChecks,
                CheatCategory.Target => _antiCheatConfig.EnableTargetChecks,
                CheatCategory.Other => true,
                _ => throw new ArgumentOutOfRangeException(nameof(category), category, null),
            };

            if (!isCategoryEnabled)
            {
                return false;
            }
```

src/Impostor.Api/Config/AntiCheatConfig.cs
```diff
@@ -6,6 +6,22 @@ public class AntiCheatConfig
...
+
+        public bool EnableMustBeHostChecks { get; set; } = true;
+
+        public bool EnableLimitChecks { get; set; } = true;
+
+        public bool EnableOwnershipChecks { get; set; } = true;
+
+        public bool EnableRoleChecks { get; set; } = true;
+
+        public bool EnableTargetChecks { get; set; } = true;
```
js6pak, line 23: Did you consider making a `Dictionary<CheatCategory, bool> EnabledCategories` or something like that instead?

js6pak (changes requested):

src/Impostor.Server/Net/Client.cs
miniduikboot: Accepted but with a different discard pattern, I don't think throwing exceptions from the API is acceptable. It may be desireable to make changes in the CheatCategory enum in future, and throwing on unknown categories could cause old plugins to break

src/Impostor.Server/Net/Client.cs
miniduikboot: You're correct, fixed.

I don't agree on renaming it back, "Exempt" is an obscure word so I think ForbidHostCheating is more clear.

src/Impostor.Api/Config/AntiCheatConfig.cs
miniduikboot: No, I didn't know how configs could be nested. A dictionary sounds annoying to set via

miniduikboot: No, I didn't know how configs could be nested. A dictionary sounds annoying to set via environment variables

src/Impostor.Server/Net/Client.cs
js6pak: >  "Exempt" is an obscure word 

Is it?

> ForbidHostCheating is more clear

It sounds more like a name for an exploit that allows you to cheat yourself to be the host.

src/Impostor.Server/Net/Client.cs
js6pak: >  could cause old plugins to break

In what way?

miniduikboot: If we ever remove an old category, but that's an API break. Considering the average plugin quality and because C# allows casting numbers to enums crashing on an invalid anticheat category seems harsh.

src/Impostor.Server/Net/Client.cs
miniduikboot: Reworked it to be tristate to accomodate the DisableHostAuthority flag. Yes, this is repurposing, but I don't think we are the first.

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -910,7 +944,7 @@ private async ValueTask<bool> HandleSetStartCounter(ClientPlayer sender, int seq
         {
             if (!sender.IsHost && startCounter != -1)
             {
-                if (await sender.Client.ReportCheatAsync(RpcCalls.MurderPlayer, "Client tried to set start counter as a non-host"))
+                if (await sender.Client.ReportCheatAsync(RpcCalls.SetStartCounter, CheatCategory.MustBeHost, "Client tried to set start counter as a non-host"))
```
AeonLucid, line 947: I assume this being `RpcCalls.MurderPlayer` before was a typo?

docs/Server-configuration.md
```diff
@@ -27,10 +27,19 @@ Impostor has an Http Server that is used by recent versions of Among Us to conne
...
+| **AllowCheatingHosts**     | `"Never"` | Configure whether hosts are allowed to cheat. "Never" forbids it, "Always" allows it. "IfRequested" allows hosts to cheat if they connect with the DisableServerAuthorityFlag set.                                  |
+| **EnableGameFlowChecks**   | `true`    | Enable checks that check if certain actions are done in the appropriate order or at the appropriate moment in the game. This includes changing cosmetics while in game or murdering too fast.                       |
+| **EnableMustBeHostChecks** | `true`    | Enables checks that check if players are the host before they can do actions that require them to be host of the game. This includes starting the game and spawning objects.                                        |
+| **EnableLimitChecks**      | `true`    | Enables checks that checks if player names have a length that is possible to set using the user interface.                                                                                                          |
```
AeonLucid, line 37: Limit description: Also specify about the duplicate color.

Also not sure if `Limit` is the best description for this AntiCheat rule but I can not think of something better.

src/Impostor.Api/Config/CheatingHostMode.cs
```diff
@@ -0,0 +1,9 @@
+namespace Impostor.Api.Config
+{
+    public enum CheatingHostMode
+    {
+        Never,
+        IfRequested,
```
AeonLucid, line 6: Some comments would be nice like it was done for `CheatCategory`.

AeonLucid (changes requested):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: I could not find a good reason for it to be `MurderPlayer`, so I corrected it

docs/Server-configuration.md
AeonLucid: Alternative suggestions:
- `NameValidations` / `ColorValidations`
- `NameLimits` / `ColorLimits`

src/Impostor.Api/Config/CheatingHostMode.cs
miniduikboot: Added documentation comments

docs/Server-configuration.md
miniduikboot: Split into Name and Color limits

AeonLucid (approved):

miniduikboot:
Thanks for reviewing. I'll squash-merge this because of the amount of fixes and reworks on this branch.

I'm ignoring @js6pak's request for changes as I've addressed his feedback and rerequested review, but received no response.

PR #563, opened 2024-01-24
Upgrade to net8.0

js6pak:


miniduikboot (commented):
I'm seeing some warnings about using obsolete formatter-based serialization and an obsolete function. Should we address these before merging?

miniduikboot (approved):

PR #565, opened 2024-01-25
Bump version in Directory.Build.props

miniduikboot:
### Description
Release build failed and the docker image claims it's the full blown version, which it isn't: it's just an RC

### Closes issues

- closes #


miniduikboot:
Doesn't work

PR #571, opened 2024-03-13
Miscellaneous Anticheat fixes

miniduikboot:
### Description

After testing with some host-only mods who can't currently use the official servers, we found a bunch of issues related to intentional desyncs triggering the anticheat and some checks being enabled unconditionally.

Marking as draft because #559 has not yet been merged, but this seems out of scope for that PR

### Closes issues

- closes various issues raised by the TOHE community: https://discord.com/channels/1094344790910455908/1217378303703252992/1217378303703252992 and https://discord.com/channels/1094344790910455908/1215692863598501918/1215692863598501918

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -804,7 +804,9 @@ namespace Impostor.Server.Net.Inner.Objects
                 }
             }
 
-            if (target == null || target.PlayerInfo.IsImpostor)
+            // Host-only mods intentionally desync players, so it may appear that one killing role (like a genuine impostor) is killing another
+            // killing role (like an sheriff). So this needs to be allowed if the host requested authority.
+            if (target == null || (target.PlayerInfo.IsImpostor && !_game.IsHostAuthoritive))
             {
                 if (await sender.Client.ReportCheatAsync(RpcCalls.CheckMurder, CheatCategory.GameFlow, "Client tried to murder invalid target"))
                 {
@@ -849,7 +851,7 @@ namespace Impostor.Server.Net.Inner.Objects
                 }
             }
 
-            if (target == null || target.PlayerInfo.IsImpostor)
+            if (target == null)
             {
                 if (await sender.Client.ReportCheatAsync(RpcCalls.MurderPlayer, CheatCategory.GameFlow, "Client tried to murder invalid target"))
                 {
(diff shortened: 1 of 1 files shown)


AeonLucid:
I can not see the discord links please provide screenshots the next time

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -647,9 +647,11 @@ private async ValueTask<bool> HandleSetName(ClientPlayer sender, string name)
 
                     if (name != expected)
                     {
-                        _logger.LogWarning($"Client sent {nameof(RpcCalls.SetName)} with incorrect name");
-                        await SetNameAsync(expected);
-                        return false;
+                        if (await sender.Client.ReportCheatAsync(RpcCalls.SetName, CheatCategory.GameFlow, "Client sent SetName with incorrect name"))
```
AeonLucid, line 650: Is this something that can never trigger legitimately? The loop above this is a bit difficult to read but I guess its to de-dupe duplicate names with a number suffix?

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -711,24 +713,30 @@ private async ValueTask<bool> HandleSetColor(ClientPlayer sender, ColorType colo
...
-                if (!RequestedColorId.Any())
+                if (RequestedColorId.Any())
                 {
-                    _logger.LogWarning($"Client sent {nameof(RpcCalls.SetColor)} for a player that didn't request it");
-                    return false;
-                }
+                    var expected = RequestedColorId.Dequeue();
 
-                var expected = RequestedColorId.Dequeue();
+                    while (Game.Players.Any(x => x.Character != null && x.Character != this && x.Character.PlayerInfo.CurrentOutfit.Color == expected))
```
AeonLucid, line 720: Technically this can infinitely loop when all ColorTypes are picked. I guess the loop is trying to assign a random new color if the player sends one that is already in use? What if it randomly assigns one that another player in the next packet wanted?

(I see that it's not a change of this PR but still)

AeonLucid (changes requested):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: I think that's correct, that should also be the behaviour of vanilla when it encounters two players with the same name. So vanilla should never trigger this

AeonLucid (approved):

miniduikboot:
Forced pushed twice:
- Rebased on top of master
- Retrigger builds because Appveyor (!) claimed my PR was unmergeable and that was causing the build to fail

As the contents have not changed since review, I'm going to merge this normally

PR #591, opened 2024-04-21
[AntiCheat] Add ValidateBroadCast check to most of the rpc calls

NikoCat233:
Add ValidateBroadCast check to most of the rpc calls
Shapeshifter role check should also be put up for checkShapeshift

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -118,6 +118,11 @@ namespace Impostor.Server.Net.Inner.Objects
                         return false;
                     }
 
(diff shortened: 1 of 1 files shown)


miniduikboot (approved):
Thanks for your PR, I noticed you missed a few RPC's so I added these as well. And while I was in there, I did some cleanups too.

NikoCat233:
Instead of removing SetInfected check, I think u should keep the check there. Once sent issue a kick

miniduikboot:
By removing the case entirely Impostor will execute the default branch, which will end up in InnerNetObject.UnregisteredCall which kicks in the category ProtocolExtension. The reason I can't do that with SyncSettings is because players still call SyncSettings but don't handle it at all anymore. But there are no cases left where SetInfected will be called, so it is safe to remove the case entirely

PR #595, opened 2024-05-05
Add SanctionReason

NikoCat233:
image

https://github.com/Impostor/Impostor/assets/139348239/8c218723-e8f8-47db-b411-a5e0396f8307

### DIFF
--- a/src/Impostor.Api.Innersloth.Generator/SourceGenerator.cs
+++ b/src/Impostor.Api.Innersloth.Generator/SourceGenerator.cs
@@ -75,6 +75,7 @@ public sealed class SourceGenerator : IIncrementalGenerator
             enumGenerator.Generate("ColorType", "Impostor.Api.Innersloth.Customization");
 
             enumGenerator.Generate("DisconnectReason", sourceName: "DisconnectReasons");
+            enumGenerator.Generate("SanctionReason", sourceName: "SanctionReasons");
             enumGenerator.Generate("GameKeywords", flags: true, underlyingType: CSharpEnumUnderlyingType.UnsignedInt);
             enumGenerator.Generate("GameOverReason", underlyingType: CSharpEnumUnderlyingType.Byte);
             enumGenerator.Generate("Platforms");
--- a/src/Impostor.Api/Innersloth/Data/enums/SanctionReasons.json
+++ b/src/Impostor.Api/Innersloth/Data/enums/SanctionReasons.json
@@ -1,0 +1,23 @@
+{
+  "None": 0,
+  "ImpersonationCeleb": 1,
+  "Spamming": 2,
+  "InappropriateNameUnsportsmanlike": 3,
+  "UnsportsmanlikeConduct": 4,
+  "ImpersonationDevelopers": 5,
+  "InappropriateChatPersonalInfo": 6,
+  "InappropriateNameDerogatory": 7,
+  "InappropriateNameNsfw": 8,
+  "Bullying": 9,
+  "CheatingHacking": 10,
+  "InappropriateChatDating": 11,
+  "WeaponizingRules": 12,
+  "RepeatOffender3": 13,
+  "SexualMisconduct": 14,
+  "Doxing": 15,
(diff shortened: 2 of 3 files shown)


NikoCat233:
To use this, DisconnectReason should be DisconnectReason.Sanctions 
and you must assign a EndsAt

src/Impostor.Api/Innersloth/Data/enums/SanctionReasons.json
miniduikboot, line 1: Did you manually create this file? It should be automatically generated using Dumpostor: https://github.com/Impostor/Dumpostor/blob/master/Dumpostor/DumpostorPlugin.cs#L90

NikoCat233: hm what is this

NikoCat233: image

NikoCat233: Oh i see, i will gen it

NikoCat233:
Now its dumped with Dumpostor

AeonLucid:
> Litterly Same with what I manully entered lol

Makes maintaining and updating it easier in the future

miniduikboot (changes requested):
This PR doesn't work correctly: the message that is sent when joining a non-existing game changed to something more confusing.

Reproduction steps:
1. Start Impostor with a build from your branch.
2. Join the non-existent game "AAAA"
3. Wait for the error message.

Before: Could not find the game you're looking for.
After: There was a problem finding a game code.\n Not found. Code: 404

I don't believe this change is intended by you, could you look into why this message changed and make sure the old message is shown again?

js6pak:
> Before: Could not find the game you're looking for. After: There was a problem finding a game code.\n Not found. Code: 404

This was caused by the game client not accepting `null` `EndsAt` values and crashing

I've pushed a fix and this is how the game displays errors now:

`new MatchmakerError(DisconnectReason.GameNotFound)`
image

`new MatchmakerError(SanctionReason.CheatingHacking, DateTimeOffset.UtcNow)`
image

`new MatchmakerError(SanctionReason.CheatingHacking, DateTimeOffset.MaxValue)`
image

src/Impostor.Api.Innersloth.Generator/SourceGenerator.cs
```diff
@@ -75,6 +75,7 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
             enumGenerator.Generate("ColorType", "Impostor.Api.Innersloth.Customization");
 
             enumGenerator.Generate("DisconnectReason", sourceName: "DisconnectReasons");
+            enumGenerator.Generate("SanctionReason", sourceName: "SanctionReasons");
```
js6pak, line 78: What's the reason for renaming it to `SanctionReason`? Both Impostor and base game are inconsistent in naming of enums, so I think we should just stick with base game's `SanctionReasons`.

src/Impostor.Api/Innersloth/Data/info.json
```diff
@@ -1,5 +1,5 @@
 {
-  "dumpostorVersion": "1.0.0",
+  "dumpostorVersion": "1.0.0-dev",
```
js6pak, line 2: This should be regenerated with release dumpostor before merging

js6pak (changes requested):
All of this is currently internal, so what's the point of this PR? Do we want to open this up for plugins?

NikoCat233:
> All of this is currently internal, so what's the point of this PR? Do we want to open this up for plugins?

Probably, im using it on my own server

NikoCat233:
I think changes is done

miniduikboot:
6pak's comments have not yet been addressed: the enum hasn't been regenerated with a release version of Dumpostor (please trigger him for that) and I still don't quite see the point of this change as it doesn't expose the API to users in any way

PR #596, opened 2024-05-26
Clean up clients that refuse to quit properly

miniduikboot:
When a player is kicked or banned from a game, the client is expected to disconnect in response. When they decide not to *or* the disconnect packet is lost, Impostor just silently disposes of the connection. This does not remove them from ClientManager, which is how I tracked this bug down.

By sending them a proper disconnect message, clients should have received *a* disconnection message before dropping their connection.

Future work should split this entire method in removing players from a game and processing kick/bans because the latter feature is not relevant for other callers of this method.


AeonLucid (approved):
Makes sense

miniduikboot:
Tested in production for 2 days, works as expected

PR #597, opened 2024-06-04
Add 2024.6.4 game version

miniduikboot:
Officials allow crossplay and after inspecting the Mono I don't see any major changes, so I'm following officials here

Keeping open for 24h, will merge afterwards


NikoCat233:
Seems like they only added some comestics

PR #598, opened 2024-06-13
Accept RPC's called on GameManager

miniduikboot:
### Description

Previously Impostor would just throw when a method was called on a GameManager, but this caused issues with client mods that did just this, even through vanilla never calls an RPC on GameManager.

This commit changes that throw statement to pass the exception to the custom message handler/anticheat instead, which is also how it works for the other NetObjects

### Closes issues

- Fixed 25 exceptions caused by various mods on my server in my logging window

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
@@ -53,12 +53,20 @@ internal abstract class InnerGameManager : InnerNetObject, IInnerGameManager
 
     public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, ClientPlayer? target, RpcCalls call, IMessageReader reader)
     {
+        // Let all logic components process this RPC. If at least one component handles it, return true.
+        var result = false;
         foreach (var logicComponent in _logicComponents)
         {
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
```diff
@@ -53,12 +53,20 @@ internal int GetGameLogicTag<T>(T logic)
 
     public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, ClientPlayer? target, RpcCalls call, IMessageReader reader)
     {
+        // Let all logic components process this RPC. If at least one component handles it, return true.
+        var result = false;
         foreach (var logicComponent in _logicComponents)
         {
-            await logicComponent.HandleRpcAsync(call, reader);
+            result |= await logicComponent.HandleRpcAsync(call, reader);
```
AeonLucid, line 60: Nifty

AeonLucid (approved):

src/Impostor.Server/Net/Inner/Objects/GameManager/InnerGameManager.cs
miniduikboot: Yeah, I tried to keep the behavior of the previous code intact and `|=` doesn't short circuit iirc

PR #599, opened 2024-06-14
Improve CompatMgr behaviour for unknown versions

miniduikboot:
### Description

Currently, Impostor will show an "Client is in an invalid state." message if a player tries to join a match where the host or the joining player is using a version unknown to Impostor.

This message confused a *lot* of people, so this PR changes the API such that if the version is not registered, the version itself is put in a compatibility group and compared, assuming no other game versions are compatible with it. This works fine in most cases, as Among Us network versions generally increase over time.

Failure on 2222.2.2 (which is ancient at this point, so likely no longer used) and future cases where Innersloth messes up (which is unlikely) is an acceptable consequence to me.

### DIFF
--- a/src/Impostor.Server/Net/Manager/CompatibilityManager.cs
+++ b/src/Impostor.Server/Net/Manager/CompatibilityManager.cs
@@ -63,6 +63,12 @@ internal class CompatibilityManager : ICompatibilityManager
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


miniduikboot:
I'd like to merge this fix tomorrow. Please review otherwise I'll roll it up into the next PR

Pietrodjaowjao:
you will need to edit it since tomorrow 2024.6.18 will release

AeonLucid (approved):

PR #603, opened 2024-07-02
ProtocolExtensions mode + setting

NikoCat233:
Simply setting this setting as true or false is not a good practice.
We can decide this based on host authority. Any mod that uses protocol extensions should have +25 flags

### DIFF
--- a/src/Impostor.Api/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Api/Config/AntiCheatConfig.cs
@@ -1,4 +1,4 @@
-﻿namespace Impostor.Api.Config
+namespace Impostor.Api.Config
 {
     public class AntiCheatConfig
     {
@@ -24,6 +24,6 @@
 
         public bool EnableTargetChecks { get; set; } = true;
 
-        public bool ForbidProtocolExtensions { get; set; } = true;
+        public ProtocolExtensionsMode AllowProtocolExtensions { get; set; } = ProtocolExtensionsMode.Never;
     }
 }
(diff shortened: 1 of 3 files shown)


NikoCat233:
I'm not good at writing documents, help

miniduikboot:
Thanks for your pull request, however I don't think this is an improvement to Impostor. When I added the modular anticheat, it included an option to allow hosts cheating if they enabled the +25 mode. The idea was that giving hosts the ability to cheat would make it possible to enable AC on mods, as they could request the host to do things that violate the anticheat. It would also make mods more easily compatible with Impostor's anticheat, making it possible to deploy an anticheat on my modded servers.

> Any mod that uses protocol extensions should have +25 flags

This has historically not been the case, and with 2024.6.18 Innersloth is making large changes to the network protocol while disabling most of it for +25 hosts. However with the amount of networking issues that are present on 2024.6.18 when +25 is in use I can't recommend its use anymore for mods unless they need the changes provided by +25, like host-only mods. I don't see this improving in future, and I predict that the next Among Us patch will again break +25 until it becomes unusable like DnsRegionInfo.

Protocol extensions however are currently triggered by 3 things: players sending unknown RPC's, players entering a vent not known to the server or players entering a vent on an unknown map. Neither are relied upon by host-only mods, as they rely on modifying the existing RPC's in the game to create a new experience and can't use RPC's that are ignored by players. Therefore this commit would benefit neither host-only mods (as they don't use ProtocolExtensions) nor all-client mods (as they don't use +25)

I'm regretting my decision to add a host cheating bypass to the modular anticheat, but I can't remove just remove it due to backwards compatibility. However I do recognize that there are use cases for clients selectively disabling (parts of) the anticheat, so I'm currently thinking about interesting ways to allow mods to disable anticheat rules they cannot comply with. If you're also interested I suggest we pick this up after the 2024.6.18 update has shipped.

PR #604, opened 2024-07-08
Add Rpc35UpdateSystem

NikoCat233:
- [ ] Basic read and write rpc calls
- [ ] Add new system types and do correct Serialize and Deserialize
- [ ] Update System Anticheat (check amount, allow messagewriter only on VentilationSystem)
- [ ] Update System events


NikoCat233:
Closed due to the fact that I never gonna finish this

PR #607, opened 2024-07-23
Refactor Start/FinishRpc to add GameData methods

miniduikboot:
### Description

Rpc's are a specific sort of GameData packets, so it makes sense to split and extract a method for sending GameData packets from these functions.

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Outgoing.cs
+++ b/src/Impostor.Server/Net/State/Game.Outgoing.cs
@@ -40,6 +40,27 @@ namespace Impostor.Server.Net.State
 
         public IMessageWriter StartRpc(uint targetNetId, RpcCalls callId, int? targetClientId = null, MessageType type = MessageType.Reliable)
         {
+            var writer = StartGameData(targetClientId, type);
+
+            writer.StartMessage(GameDataTag.RpcFlag);
+            writer.WritePacked(targetNetId);
+            writer.Write((byte)callId);
+
+            return writer;
+        }
+
+        public ValueTask FinishRpcAsync(IMessageWriter writer, int? targetClientId = null)
+        {
+            writer.EndMessage();
+            return FinishGameDataAsync(writer, targetClientId);
+        }
+
+        /// <summary>Start a GameData(To) message.</summary>
+        /// <param name="targetClientId">The client to target if needed, `null` otherwise.</param>
+        /// <param name="type">The type of message to send, defaults to reliable.</param>
+        /// <returns>MessageWriter that should be handed back to <see cref="FinishGameDataAsync"/>.</returns>
+        private IMessageWriter StartGameData(int? targetClientId = null, MessageType type = MessageType.Reliable)
+        {
             var writer = MessageWriter.Get(type);
 
             if (targetClientId == null || targetClientId < 0)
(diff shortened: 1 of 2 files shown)


js6pak:
`Game.SyncSettingsAsync` seems to be the only place besides rpcs where we craft a GameData message, do we want to use `StartGameData` or even `SendObjectSpawn` there?

src/Impostor.Server/Net/State/Game.Outgoing.cs
```diff
@@ -39,6 +39,27 @@ public async ValueTask SendToAsync(IMessageWriter writer, int id)
...
+        {
+            writer.EndMessage();
+            return FinishGameDataAsync(writer, targetClientId);
+        }
+
+        /// <summary>Start a GameData(To) message.</summary>
+        /// <param name="targetClientId">The client to target if needed, `null` otherwise.</param>
+        /// <param name="type">The type of message to send, defaults to reliable.</param>
+        /// <returns>MessageWriter that should be handed back to <see cref="FinishGameDataAsync"/>.</returns>
+        private IMessageWriter StartGameData(int? targetClientId = null, MessageType type = MessageType.Reliable)
```
js6pak, line 62: Why not public?

js6pak (changes requested):

src/Impostor.Server/Net/State/Game.Outgoing.cs
miniduikboot: Good question. I guess we can add this to the public API like the RPC variant, although it removes the option to add a nicer and safer RAII-style API later.

If that's an acceptable tradeoff I guess I can make it public in this PR

miniduikboot:
> `Game.SyncSettingsAsync` seems to be the only place besides rpcs where we craft a GameData message, do we want to use `StartGameData` or even `SendObjectSpawn` there?

We should probably use the SendObjectData method from #616 . Out of scope for this PR, do you want to craft a followup PR later?

src/Impostor.Server/Net/State/Game.Outgoing.cs
miniduikboot: Note that StartRpc is annotated with the following API:

```
/// <remarks>Use with caution, recommended only for advanced developers.</remarks>
```
GameData is even harder to approach than Rpc. I don't see a plugin use case that would benefit from sending a Spawn/Despawn/SceneChange/Ready packet manually. If we want to support this, we should add high level API's for this instead instead of adding a footgun like this.

js6pak: >  I don't see a plugin use case that would benefit from sending a Spawn/Despawn/SceneChange/Ready packet manually

Well the thing is that you wouldn't send those, but it's possible you might want custom GameData packets for whatever reason: https://github.com/NuclearPowered/Reactor/blob/21c0e61f57ec40677ec769167191f53336367453/Reactor/Networking/Extensions/ClientExtensions.cs#L11-L34

> although it removes the option to add a nicer and safer RAII-style API later.

I would say we can have both. I'm thinking we can have a big PR later that also adds stuff like `StartMessageScope` extension to `MessageWriter` and move all usages to it.

js6pak (approved):

PR #612, opened 2024-07-23
Add Sequence ID and Dirty tracking to PlayerOutfit

miniduikboot:
### Description

2024.6.18 adds SequenceId's to player outfits. They're currently unused and that seems to work Just Fine as long as only one player is setting cosmetics. This will probably need a follow-up PR in future.

As we need to send over the new outfit when a change occurs, implement change tracking. To prevent unnecessary code duplication, this is done using a helper method.

### DIFF
--- a/src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
+++ b/src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
@@ -2,6 +2,8 @@ namespace Impostor.Api.Innersloth.Customization
 {
     public class PlayerOutfit
     {
+        public string PlayerName { get; internal set; } = string.Empty;
+
         public ColorType Color { get; internal set; } = (ColorType)(-1);
 
         public string HatId { get; internal set; } = "missing";
@@ -14,7 +16,15 @@ namespace Impostor.Api.Innersloth.Customization
 
         public string NamePlateId { get; internal set; } = "missing";
 
-        public string PlayerName { get; internal set; } = "missing";
+        private byte HatSequenceId { get; set; } = 0;
+
+        private byte PetSequenceId { get; set; } = 0;
+
+        private byte SkinSequenceId { get; set; } = 0;
+
+        private byte VisorSequenceId { get; set; } = 0;
+
+        private byte NamePlateSequenceId { get; set; } = 0;
 
         public bool IsIncomplete
         {
@@ -38,6 +48,11 @@ namespace Impostor.Api.Innersloth.Customization
             writer.Write(SkinId);
             writer.Write(VisorId);
             writer.Write(NamePlateId);
+            writer.Write(HatSequenceId);
+            writer.Write(PetSequenceId);
+            writer.Write(SkinSequenceId);
+            writer.Write(VisorSequenceId);
+            writer.Write(NamePlateSequenceId);
         }
 
         public void Deserialize(IMessageReader reader)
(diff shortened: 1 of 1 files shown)


src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
```diff
@@ -2,19 +2,67 @@ namespace Impostor.Api.Innersloth.Customization
 {
     public class PlayerOutfit
     {
-        public ColorType Color { get; internal set; } = (ColorType)(-1);
+        private string _playerName = string.Empty;
+        private ColorType _color = (ColorType)(-1);
```
NikoCat233, line 6: InnerSloth assigns 0 red here for initial data

NikoCat233 (changes requested):

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
miniduikboot: Hmm, fair point, this is a bit ugly indeed. We can't however just set this to 0 as that'd interfere with the used color anticheat rule. What do you think about making this nullable instead, and serializing this as 0?

Considering that it was set to `-1` already in the old version, do you think this commit should be modified to add this change or should this be a separate PR?

NikoCat233: i think putting -1 is better than what innersloth does

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
```diff
@@ -2,19 +2,67 @@ namespace Impostor.Api.Innersloth.Customization
...
+        public byte PetSequenceId { get; internal set; } = 0;
 
-        public string VisorId { get; internal set; } = "missing";
+        public byte SkinSequenceId { get; internal set; } = 0;
 
-        public string NamePlateId { get; internal set; } = "missing";
+        public byte VisorSequenceId { get; internal set; } = 0;
 
-        public string PlayerName { get; internal set; } = "missing";
+        public byte NamePlateSequenceId { get; internal set; } = 0;
```
js6pak, line 65: Should we be exposing sequence ids in the public api?

js6pak (changes requested):

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
miniduikboot: Innersloth assigns -1 here too (see NetworkedPlayerInfo.PlayerOutfit.ColorId). Resolving this discussion.

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
miniduikboot: Probably not, although we should add an option to set them using Impostor.Server.

miniduikboot: I've made them private for now

miniduikboot:
I've changed the access on the cosmetic setters to internal like they were before and noted on my todo list that we need to re-evaluate the SetHatAsync etc API's at a later point. I don't think we can deprecate those right now because +25 exists, but it needs a rework regardless

js6pak (approved):

PR #614, opened 2024-07-23
Spawn PlayerInfo's if host is not authoritive 

miniduikboot:
### Description

2024.6.18 by default doesn't spawn PlayerInfo objects, rather it expects the server to do so. It reserves the special range of 10000 and above for this to prevent client and server creating objects at the same time.

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Data.cs
+++ b/src/Impostor.Server/Net/State/Game.Data.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Impostor.Api;
 using Impostor.Api.Innersloth;
@@ -30,6 +31,16 @@ namespace Impostor.Server.Net.State
         /// </summary>
         private const int CurrentClient = -3;
 
+        /// <summary>
+        ///     Used to list objects that are managed by the game server.
+        /// </summary>
+        private const int ServerOwned = -4;
+
+        /// <summary>
+        ///     The first NetId that is considered as a server owned Network ID that the client will not allocate by default.
+        /// </summary>
+        private const int MinServerNetId = 100000;
+
         private static readonly Dictionary<uint, Type> SpawnableObjects = new()
         {
             [0] = typeof(InnerSkeldShipStatus),
@@ -47,9 +58,13 @@ namespace Impostor.Server.Net.State
             [13] = typeof(InnerFungleShipStatus),
         };
 
+        private static readonly Dictionary<Type, uint> SpawnableObjectIds = SpawnableObjects.ToDictionary((i) => i.Value, (i) => i.Key);
+
         private readonly List<InnerNetObject> _allObjects = new List<InnerNetObject>();
 
         private readonly Dictionary<uint, InnerNetObject> _allObjectsFast = new Dictionary<uint, InnerNetObject>();
+
+        private uint _nextNetId = MinServerNetId;
 
         public T? FindObjectByNetId<T>(uint netId)
             where T : IInnerNetObject
@@ -255,6 +270,10 @@ namespace Impostor.Server.Net.State
                         sender.Scene = scene;
 
                         _logger.LogTrace("> Scene {0} to {1}", clientId, sender.Scene);
+
+                        await SyncServerObjectsAsync(sender);
+                        await SpawnPlayerInfoAsync(sender);
+
                         break;
                     }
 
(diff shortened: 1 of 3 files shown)


src/Impostor.Server/Net/State/Game.Outgoing.cs
```diff
@@ -100,5 +120,27 @@ private void WriteWaitForHostMessage(IMessageWriter message, bool clear, IClient
...
+            writer.WritePacked(obj.OwnerId);
+            writer.Write((byte)obj.SpawnFlags);
+
+            var components = obj.GetComponentsInChildren<InnerNetObject>();
+            writer.WritePacked(components.Count);
+            foreach (var comp in components)
+            {
+                writer.WritePacked(obj.NetId);
+                writer.StartMessage(1);
+                comp.SerializeAsync(writer, true);
```
js6pak, line 138:
```suggestion
                await comp.SerializeAsync(writer, true);
```

src/Impostor.Server/Net/State/Game.Data.cs
```diff
@@ -442,6 +448,41 @@ private async ValueTask OnDestroyAsync(InnerNetObject netObj)
...
         }
 
+        private async ValueTask CheckPlayerInfos(ClientPlayer sender)
+        {
+            // Sync all server-owned objects
+            foreach (var obj in _allObjectsFast.Values)
+            {
+                if (obj.OwnerId == ServerOwned)
+                {
+                    _logger.LogTrace("Sharing {Type} {NetId}", obj.GetType(), obj.NetId);
```
js6pak, line 458: "Sharing"?

src/Impostor.Server/Net/State/Game.Outgoing.cs
js6pak, line 138: nit:
```suggestion
            foreach (var component in components)
            {
                writer.WritePacked(obj.NetId);
                writer.StartMessage(1);
                component.SerializeAsync(writer, true);
```

js6pak (changes requested):

src/Impostor.Server/Net/State/Game.Outgoing.cs
```diff
@@ -100,5 +120,27 @@ private void WriteWaitForHostMessage(IMessageWriter message, bool clear, IClient
...
+                writer.EndMessage();
+            }
+
+            writer.EndMessage();
+            return FinishGameDataAsync(writer, targetClientId);
```
js6pak, line 143: Returning the task instead async+awaiting here means that the writer will get disposed after this method returns and before it's awaited.

js6pak (changes requested):

src/Impostor.Server/Net/State/Game.Outgoing.cs
js6pak: The same issue is in `OnQueryPlatformIds` :P

src/Impostor.Server/Net/State/Game.Data.cs
```diff
@@ -442,6 +448,41 @@ private async ValueTask OnDestroyAsync(InnerNetObject netObj)
...
+                playerInfo.NetId = _nextNetId++;
+                playerInfo.OwnerId = ServerOwned;
+                playerInfo.ClientId = sender.Client.Id;
+                playerInfo.PlayerId = GameNet.GameData.GetNextAvailablePlayerId();
+
+                if (!AddNetObject(playerInfo))
+                {
+                    _logger.LogTrace("Couldn't spawn PlayerInfo");
+                    playerInfo.NetId = uint.MaxValue;
+                    return;
```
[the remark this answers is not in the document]
miniduikboot: This method is fallible, so I want to catch when it fails. It shouldn't, as hosts shouldn't allocate PlayerInfos in this range. This error message should be more severe than Trace, bumped it up to Error. 

The client will disconnect if it doesn't get a PlayerInfo after a few seconds, so additional handling doesn't seem necessary for now.

src/Impostor.Server/Net/State/Game.Outgoing.cs
```diff
@@ -100,5 +120,27 @@ private void WriteWaitForHostMessage(IMessageWriter message, bool clear, IClient
         {
             Message12WaitForHostS2C.Serialize(message, clear, Code, player.Client.Id);
         }
+
+        private ValueTask SendObjectSpawn(InnerNetObject obj, int? targetClientId = null)
```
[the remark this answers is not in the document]
miniduikboot: Game.Outgoing.cs contains 3 groups of methods:
- Senders (SendToAllAsync, SendToAllExceptAsync, SendToAsync)
- Start/Finish based API's (StartRpc, FinishRpc and the new GameData variants)
- Writers where you need to get your own writer and pick your own sender (Write*Message)

SendObjectSpawn is slightly different than the existing groups as it serializes a message like a writer, but relies on a Start/Finish API instead of letting a user pick their own.

As it doesn't modify Impostor's internal state, I don't think calling it SpawnAsync would be accurate.

src/Impostor.Server/Net/State/Game.Outgoing.cs
miniduikboot: ... what a footgun, why don't static analyzers pick up unawaited futures?

src/Impostor.Server/Net/State/Game.Data.cs
miniduikboot: -> Syncing

src/Impostor.Server/Net/State/Game.Outgoing.cs
miniduikboot: Fixed, but I'd like to see an static analyzer chek for stuff like this. Do you know if that exists?

I'd not be surprised if this exists in at least 5 other places

src/Impostor.Server/Net/State/Game.Outgoing.cs
js6pak: You would get [`CS4014`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014) if `SendObjectSpawn` was async. I guess this is yet another reason to avoid tail returning tasks vs just awaiting them.

js6pak (approved):

miniduikboot:
#607, which was a dependency for this PR has been merged. Rebasing and undrafting.

PR #616, opened 2024-07-23
Sync dirty objects

miniduikboot:
### Description

When a PlayerInfo object is changed, it needs to be synced over. This commit does that, as well as setting the IsDirty property in a few cases where this is necessary


miniduikboot:
It looks like officials don't send NetworkedPlayerInfo update packets.

Closing this PR.

PR #617, opened 2024-07-23
Sync tasks

miniduikboot:
As tasks used to be part of the GameData, 2024.6.18 has moved them to
the PlayerInfo. Because they are now called on the PlayerInfo of the
correct player, the playerId field in the RPC is no longer necessary.


miniduikboot:
Retargeted to -prepare so it is ready for review now

js6pak (approved):

PR #618, opened 2024-07-23
Set RoleWhenAlive when set to a Ghost role

miniduikboot:
### Description

2024.6.18 added a field RoleWhenAlive that sets the old role of players. Populate this when a player is set to a ghost role


js6pak:
What about death reason?

miniduikboot:
Completely forgot about that. We should probably set the death reason in #621 and default it to a new enum value "Unknown".

miniduikboot:
Rebased and added GuardianAngel as a Ghost role.

DeathReason doesn't need to be set, it should already be set before a ghost role is assigned (Either when exiling, see #621, or when CheckMurder/MurderPlayer is called)

js6pak (approved):

PR #620, opened 2024-07-23
Reset PlayerInfo on PlayerControl spawn

miniduikboot:
When (re)spawning players get a new player control netobject, which is
the perfect moment to discharge them from their former duties and reset
their PlayerInfo to a default state: no role, no tasks and not dead.


miniduikboot:
No longer necessary because #616 is closed

PR #621, opened 2024-07-23
Let hosts handle exiles

miniduikboot:
### Description

Previously Impostor liked to kill exiled players server side to raise an event. This is quite early, as the host will SetRole them to a ghost role later.

However as PlayerInfo is server-owned by default in 2024.6.18, it starts killing players client side. This caused meetings to immediately conclude if the final impostor was voted out, ruining the cutscene.

So this commit makes Impostor wait for hosts to actually kill the player using SetRole.

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
@@ -289,7 +289,7 @@ namespace Impostor.Server.Net.Inner.Objects
 
             if (exiled != null)
             {
-                exiled.Die(DeathReason.Exile);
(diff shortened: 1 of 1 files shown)


miniduikboot:
It's not going extinct, it's just that in 2024.6.18 the server is now actually able to kill players, which it couldn't do previously as a server-side death was previously never synced back to players. That's why this call has to go, as #616 is now marking the PlayerInfo as dirty on death.

js6pak:
> It's not going extinct

It is because this is the only place where it's used (besides the `ExileAsync` api)

miniduikboot:
Changed to set the LastDeathReason property. Die should be kept as-is as it is responsible for marking players as dead on CheckMurder/MurderPlayer

js6pak (approved):

PR #625, opened 2024-07-23
Add compatibility option to allow host authority

miniduikboot:
### Description

There are various bugs in the 2024.6.18 implementation of host-authoritive multiplayer such that I can't recommend players to use it. Fixing these issues on the server side is impossible, and I can't smell remotely if clients have applied the necessary fixes. The currently known bugs include at least:

- Unable to join more than 8 players in a room
- Unable for new players to join if host switched from host authoritive to server authoritive
- Host is unable to kill if it is in a host authoritive game while it thinks it is in server authoritive mode

If Innersloth fixes this mode in future game patches, this option may be reconsidered.

### DIFF
--- a/docs/Server-configuration.md
+++ b/docs/Server-configuration.md
@@ -43,12 +43,13 @@ Impostor has an Anticheat that makes it possible to kick cheaters from games aut
 
 ### Compatibility
 
-Impostor has two compatibility options which allow some extra flexibility but may not work properly. Enabling either of these options is not recommended.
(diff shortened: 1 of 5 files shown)


NikoCat233 (changes requested):
Seems like u didnt add this setting in config-full.json

miniduikboot:
config-full.json no longer exists as of this commit, it was renamed to full-config-example.json in #624

js6pak (approved):

PR #626, opened 2024-07-23
Add extra debug log statements

miniduikboot:
### Description

To help debugging Impostor, this commit prints a message every time a
GameData message arrives, and one more if it's an Rpc or a Data update.
Most other interesting Gamedata tags already have log statements
attached, but these two didn't as they're very frequent.

They only show up when the Debug verbosity is used, which production
servers shouldn't have enabled.


NikoCat233 (approved):

miniduikboot:
I'm no longer interested in pushing this change forward

PR #627, opened 2024-07-23
Add canOverrideRole param to SetRole

miniduikboot:
### Description

This is currently unused by Innersloth

### Closes issues

- closes #

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -383,7 +383,7 @@ namespace Impostor.Server.Net.Inner.Objects
                         return false;
                     }
 
-                    Rpc44SetRole.Deserialize(reader, out var role);
+                    Rpc44SetRole.Deserialize(reader, out var role, out var _);
                     PlayerInfo.RoleType = role;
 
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -383,7 +383,7 @@ public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, Client
                         return false;
                     }
 
-                    Rpc44SetRole.Deserialize(reader, out var role);
+                    Rpc44SetRole.Deserialize(reader, out var role, out var _);
```
NikoCat233, line 386: by then, should we add a anticheat here to reject double set alive role for a client?

And we may prevent setting a player to ghost at round 1 since that will blackout all clients (have no idea how to do this)

NikoCat233 (changes requested):

js6pak (approved):

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
miniduikboot: That seems out of scope for this PR, I've made a note in my todo list of future PR's

NikoCat233 (approved):

PR #628, opened 2024-07-23
Add Message24EndGameHostMigration serialization

miniduikboot:
### Description

Looks like Innersloth added this but hasn't actually used this. Weird..

When sent it makes the new host perform a full rejoin, which gums up Impostor's spawning logic. Fixing that causes the client to gum itself up. Sigh.

After some testing it seems like host migration still works fine, so this investigation wasn't needed. Committing for posterity.

### DIFF
--- a/src/Impostor.Api/Net/Messages/S2C/Message24EndGameHostMigrationS2C.cs
+++ b/src/Impostor.Api/Net/Messages/S2C/Message24EndGameHostMigrationS2C.cs
@@ -1,0 +1,21 @@
+using Impostor.Api.Games;
+
+namespace Impostor.Api.Net.Messages.S2C
+{
+    public class Message24EndGameHostMigrationS2C
+    {
+        public static void Serialize(IMessageWriter writer, GameCode gameCode, int hostClientId)
+        {
+            writer.StartMessage(MessageFlags.EndGameHostMigration);
+            gameCode.Serialize(writer);
+            writer.Write(hostClientId);
+            writer.EndMessage();
+        }
+
+        public static void Deserialize(IMessageReader reader, out GameCode gameCode, out int hostClientId)
+        {
+            gameCode = reader.ReadInt32();
+            hostClientId = reader.ReadInt32();
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Impostor.Api/Net/Messages/S2C/Message24EndGameHostMigrationS2C.cs
```diff
@@ -0,0 +1,20 @@
...
+            writer.Write(hostClientId);
+            writer.EndMessage();
+        }
+
+        public static void Deserialize(IMessageReader reader)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
```
js6pak, line 20:
```suggestion
using Impostor.Api.Games;

namespace Impostor.Api.Net.Messages.S2C
{
    public class Message24EndGameHostMigrationS2C
    {
        public static void Serialize(IMessageWriter writer, GameCode gameCode, int hostClientId)
        {
            writer.StartMessage(MessageFlags.EndGameHostMigration);
            gameCode.Serialize(writer);
            writer.Write(hostClientId);
            writer.EndMessage();
        }

        public static void Deserialize(IMessageReader reader, out GameCode gameCode, out int hostClientId)
        {
            gameCode = reader.ReadInt32();
            hostClientId = reader.ReadInt32();
        }
    }
}
```

js6pak (changes requested):

src/Impostor.Api/Net/Messages/S2C/Message24EndGameHostMigrationS2C.cs
miniduikboot: What a terrible diff, did you just change the type of gameCode?

js6pak: That and I also implemented `Deserialize`, because might as well.

miniduikboot: Applied suggestion

js6pak (approved):

PR #629, opened 2024-07-23
Don't DC the client if they send an unknown flag

miniduikboot:
### Description

This item is only used in a LogTrace call, so it seems excessive to kick a player on servers just for a log statement

### DIFF
--- a/src/Impostor.Api/Net/Messages/MessageFlags.cs
+++ b/src/Impostor.Api/Net/Messages/MessageFlags.cs
@@ -41,7 +41,7 @@ namespace Impostor.Api.Net.Messages
 
         public static string FlagToString(byte flag)
         {
-            return FlagCache[flag];
+            return FlagCache.TryGetValue(flag, out var res) ? res : $"Unknown Flag {flag}";
         }
     }
 }


NikoCat233 (approved):

src/Impostor.Api/Net/Messages/MessageFlags.cs
```diff
@@ -41,7 +41,14 @@ static MessageFlags()
...
         {
-            return FlagCache[flag];
+            if (FlagCache.TryGetValue(flag, out var res))
+            {
+                return res;
+            }
+            else
+            {
+                return $"Unknown Flag {flag}";
+            }
```
js6pak, line 51: nit:
```suggestion
            return FlagCache.TryGetValue(flag, out var res) ? res : $"Unknown Flag {flag}";
```

js6pak (approved):

src/Impostor.Api/Net/Messages/MessageFlags.cs
miniduikboot: nice golf, applied it

PR #631, opened 2024-07-28
Serialize Color -1 as 0 in PlayerOutfit

miniduikboot:
### Description

Official servers send 0 instead of -1 to prevent some client issues.

### DIFF
--- a/src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
+++ b/src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
@@ -2,8 +2,13 @@ namespace Impostor.Api.Innersloth.Customization
 {
     public class PlayerOutfit
     {
+        // Magic value that is assigned by default when the outfit is initially created. This is not
+        // `null` because existing API already exposed this magic value. It's also not 0 (which is
+        // the value exposed when serializing) because SetColor anticheat checks exist.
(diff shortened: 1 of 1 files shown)


js6pak:
Can you add a comment explaining why this is needed?
Also we should probably make a const `InvalidColor` or something for the `-1` value to avoid the magic number.

miniduikboot:
Added an UnknownColor constant and additional justification

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
```diff
@@ -64,23 +69,25 @@ public string NamePlateId
...
-                if (!string.IsNullOrEmpty(PlayerName) && Color != (ColorType)(-1) && HatId != "missing" && PetId != "missing" && SkinId != "missing" && VisorId != "missing")
-                {
-                    return NamePlateId == "missing";
-                }
-
-                return true;
-            }
-        }
+        /// <summary>
+        /// Gets a value indicating whether a player outfit is complete or whether it is needs to receive additional cosmetics.
```
js6pak, line 73:
```suggestion
        /// Gets a value indicating whether a player outfit is complete or whether it needs to receive additional cosmetics.
```

js6pak (approved):

PR #633, opened 2024-08-02
Set PlayerLevel in PlayerInfo on RPC SetLevel

miniduikboot:
Previously this used to be done by the host, but this is now the server's responsibility.

I'm unable to test this, but I'm pretty certain this would work on paper. Marking as draft until someone verifies this.

Test Plan:
- Probably merge this patch on the octopus branch to get a working Impostor
- Do not use any mods, most notably do not use any mod that enables host-authoritive networking.
- Get an Among Us account with at least level 2
- Join a game with at least 4 players
- Start a meeting
- Observe that with this patch, the player level is set correctly and without this patch it is not.

For people that attempt to be smart and UnityExplorer their way out of this: Among Us offsets levels by 1, so player level 0 will show as level 1.

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -251,7 +251,7 @@ namespace Impostor.Server.Net.Inner.Objects
                     }
 
                     Rpc38SetLevel.Deserialize(reader, out var level);
-                    return true;
+                    return await HandleSetLevel(sender, level);
                 }
 
                 case RpcCalls.ReportDeadBody:
@@ -830,6 +830,19 @@ namespace Impostor.Server.Net.Inner.Objects
             }
 
             PlayerInfo.CurrentOutfit.NamePlateId = skin;
+
+            return true;
+        }
+
+        private async ValueTask<bool> HandleSetLevel(ClientPlayer sender, uint level)
+        {
+            if (Game.GameState == GameStates.Started &&
+                await sender.Client.ReportCheatAsync(RpcCalls.SetLevel, CheatCategory.GameFlow, "Client tried to set level while not in game"))
+            {
+                return false;
+            }
+
+            PlayerInfo.PlayerLevel = level;
 
             return true;
         }


NikoCat233:
Player level is first set in client data from handshake.
Server should use that instead of trusting the rpc

miniduikboot:
Both should have the same information and as we can validate neither, I went for the road of least resistance.

Difference is that this one could be sent by other players and AC would only block it if enabled.

Pietrodjaowjao:
thats more related to cosmetics, not gameflow i think

miniduikboot:
Pietro, not sure what you're commenting on. If it's the chosen cheat category: the rest of the cosmetics rpc's also use this category and there is no separate cheat category for cosmetics. Maybe there should be, but that's out of scope for this PR.

miniduikboot:
> Player level is first set in client data from handshake. Server should use that instead of simply trusting the rpc

Sorry, I have limited access to my PC at the moment, so I didn't verify your findings when I posted my initial comment. I can't find this code. I see that the server is setting the PUID, friendcode and level in Message07JoinedGameS2C, but that's the server setting data on the client. And the server has no clue what the level should be, so it just writes `1` which corresponds with level 2.

Therefore I don't think there is an alternative to this patch.

Pietrodjaowjao:
> Pietro, not sure what you're commenting on. If it's the chosen cheat category: the rest of the cosmetics rpc's also use this category and there is no separate cheat category for cosmetics. Maybe there should be, but that's out of scope for this PR.

Yeah, im talking about the cheat category, i feel like adding a cosmetic category would be more precise than GameFlow

TianMengLucky:
> Player level is first set in client data from handshake. Server should use that instead of simply trusting the rpc
> 
> your patch should be working

Level is unable to cause destructive cheating and has little impact

miniduikboot:
Currently it has the same category as other checks like this, Impostor also checks when SetHat/Nameplate/etc are called during the game. I'd not be opposed to moving this to a new category but that was not the goal of this PR.

I'm also slightly disappointed by the amount of comments a trivial PR like this have received compared to the other pull requests that are necessary for full 6.18 support. Should I have split these up even further for the peanut gallery to review them?

PR #634, opened 2024-08-03
Add Rpc44SetRole To Changing NetworkedPlayerInfo.IsDead

Dolly1016:
### Description

Some mod requires flexible change of player state.
Currently, there is no way to reset NetworkedPlayerInfo.IsDead to False.
This PR adds means of reviving to Impostor.


Dolly1016:
I realize there is another better way, so I'm taking it down.
This PR is bad for anti-cheat.

PR #638, opened 2024-08-09
Change config.json

TianMengLucky:
### Description
The existence of two configuration files is easily misleading to the user, this is a simpler way to modify it


TianMengLucky:
Here's another https://github.com/TianMengLucky/Impostor/tree/Changeconfig2

miniduikboot (approved):
Thanks for your PR, I'll check later if the settings in the config file match the ones in this config, then I'll merge it. I doubt that we're enabling the Game Recorder by default

TianMengLucky:
> Thanks for your PR, I'll check later if the settings in the config file match the ones in this config, then I'll merge it. I doubt that we're enabling the Game Recorder by default

Changed default should be false

PR #642, opened 2024-08-21
Implementing SerializeAsync and DeserializeAsync for InnerPlayerPhysics

TianMengLucky:
### Description

Its implementation itself is a no-op and should not throw an unimplemented error

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
@@ -27,12 +27,12 @@ namespace Impostor.Server.Net.Inner.Objects.Components
 
         public override ValueTask<bool> SerializeAsync(IMessageWriter writer, bool initialState)
         {
-            throw new NotImplementedException();
+            return new ValueTask<bool>(false);
         }
 
         public override ValueTask DeserializeAsync(IClientPlayer sender, IClientPlayer? target, IMessageReader reader, bool initialState)
         {
-            throw new NotImplementedException();
+            return ValueTask.CompletedTask;
         }
 
         public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, ClientPlayer? target, RpcCalls call, IMessageReader reader)


src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
```diff
@@ -27,12 +27,12 @@ public InnerPlayerPhysics(ICustomMessageManager<ICustomRpc> customMessageManager
 
         public override ValueTask<bool> SerializeAsync(IMessageWriter writer, bool initialState)
         {
-            throw new NotImplementedException();
+            return new ValueTask<bool>(false);
```
miniduikboot, line 30: Please keep it the way it was, Impostor rarely serializes network objects (InnerPlayerInfo is the exception here) so throwing when someone attempts to serialize this object makes sense. Other classes like InnerLobbyBehavior, InnerPlayerControl, InnerMeetingHud also throw on serialize.

src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
```diff
@@ -27,12 +27,12 @@ public InnerPlayerPhysics(ICustomMessageManager<ICustomRpc> customMessageManager
...
         }
 
         public override ValueTask DeserializeAsync(IClientPlayer sender, IClientPlayer? target, IMessageReader reader, bool initialState)
         {
-            throw new NotImplementedException();
+            return ValueTask.CompletedTask;
```
miniduikboot, line 35: Does it make sense to add an anticheat check here? Vanilla doesn't send out these kinds of packets

miniduikboot (changes requested):
In principle I'm in favor of this change but I think we should limit ourselves to DeserializeAsync calls: only these can be triggered by Data packets received from clients. SerializeAsync is only triggered when spawning an object server side.

Does it also make sense to remove the `throw` statement in other network objects? Right now we have one mod that wants Impostor to deserialize a random network object for reasons that are unknown to me, next week there is another mod that is deserializing another random network object.

src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
TianMengLucky: Perhaps we are supposed to add a check to set whether or not to run Vanlia's unused correlation

src/Impostor.Server/Net/Inner/Objects/Components/InnerPlayerPhysics.cs
TianMengLucky: I think you should use Logger to prompt it instead of throwing the

PR #644, opened 2024-08-22
Fix null reference when +25 host despawn playerinfo before server could

NikoCat233:
image
+25 protocol, sometimes player disconnects and host sent despawn playerinfo before server could remove it from dictionary


miniduikboot (approved):
Oops, sorry about that. I should investigate if it is possible to add some static analysis in future to stop these kinds of issues

miniduikboot:
Shit, I accidentally merged this without checking if the commit message is useful. Shouldn't merge things early in the morning q.q

Niko, can you please put the information you put in the PR title also in your commit? Ideally you write a good commit message, then Github autofills the PR based on your commit

PR #645, opened 2024-08-22
Fix cheating disconnect message

js6pak:
DisconnectReason.Hacking has been broken in base game for a long time so replace it with a custom message. Also add a random support code for easier debugging in case of false positives.

Current:
image

This PR:
image

### DIFF
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -86,9 +86,14 @@ namespace Impostor.Server.Net
 
             _logger.LogWarning("Client {Name} ({Id}) was caught cheating: [{SupportCode}] [{Context}-{Category}] {Message}", Name, Id, supportCode, context.Name, category, message);
 
-            if (_antiCheatConfig.BanIpFromGame)
+            if (Player is { } player)
             {
-                Player?.Game.BanIp(Connection.EndPoint.Address);
+                if (_antiCheatConfig.BanIpFromGame)
+                {
+                    player.Game.BanIp(Connection.EndPoint.Address);
+                }
+
+                await player.Game.HandleRemovePlayer(Id, DisconnectReason.Hacking);
             }
 
             var disconnectMessage =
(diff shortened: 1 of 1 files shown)


NikoCat233:
the problem is this disconnect message is always shown as kicked to other players

NikoCat233:
public async ValueTask RemoveAsync(DisconnectReason reason, string custom = "")
        {
            await Game.HandleRemovePlayer(Client.Id, reason);
            await Client.DisconnectAsync(reason, custom);
        }

js6pak:
For the disconnect notification shown to other players we've got just these 4 options. Ideally we'd use Banned/Kicked but those specify the host as the one who bans/kicks so it doesn't really make sense for us.

> the problem is this disconnect message is always shown as kicked to other players

Currently, the "left the game" message is shown, so I don't really know what you mean?

`DisconnectReason.ExitGame` (the default in Impostor)
image

`DisconnectReason.Banned`
image

`DisconnectReason.Kicked`
image

any other `DisconnectReason`
image

miniduikboot (approved):
lgtm

NikoCat233:
image
I cant replicate what you have shown in your pictures

js6pak:
> If a player is kicked by Impostor for anticheat purpose
	I expect server telling others that "It is banned by AntiCheat", not "It exited the game itself"

As far as I know, those 4 messages I mentioned are the only ones we can use on vanilla so that wouldn't be possible.

> Also for other DisconnectReasons like custom "Reliable packet was ****"
	We should also tell others the correct Disconnect Reasons instead of the deafult ExitGame

We could indeed use the "left the game due to error" for those, but that would be a different PR.

> Currently My region is using this function to broadcast correct Disconnect Messages to other players

Are you sure that's the case on vanilla clients? Can you show a disconnect notification message different from the ones I mentioned?

NikoCat233:
My opinion is we should send the exact disconnect reasons to everyone like innersloth does
no matter what clients can actually see

PR #647, opened 2024-08-23
Fix RemovePlayer reason on internal disconnects

js6pak:
Previously all disconnections were broadcasted as `DisconnectReason.ExitGame` regardless of whether the disconnect was initiated by us or the client. This PR makes it so our disconnects use `DisconnectReason.Error` instead.

If in the future base game supports more `DisconnectReason`s in the notification message we could use more specific reasons using `HazelInternalErrors` provided by `NetworkConnection.OnInternalDisconnect`.

pressing "Leave Game":
image

network timeout:
image

cc @NikoCat233

### DIFF
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -348,7 +348,9 @@ namespace Impostor.Server.Net
             {
                 if (Player != null)
                 {
-                    await Player.Game.HandleRemovePlayer(Id, DisconnectReason.ExitGame);
+                    // The client never sends over the real disconnect reason so we always assume ExitGame
+                    var isRemote = reason == "The remote sent a disconnect request";
+                    await Player.Game.HandleRemovePlayer(Id, isRemote ? DisconnectReason.ExitGame : DisconnectReason.Error);
                 }
             }
             catch (Exception ex)


NikoCat233 (approved):

src/Impostor.Server/Net/Client.cs
```diff
@@ -348,7 +348,9 @@ public override async ValueTask HandleDisconnectAsync(string reason)
             {
                 if (Player != null)
                 {
-                    await Player.Game.HandleRemovePlayer(Id, DisconnectReason.ExitGame);
+                    // The client never sends over the real disconnect reason so we always assume ExitGame
+                    var isRemote = reason == "The remote sent a disconnect request";
```
miniduikboot, line 352: Magic string, consider moving it into a special string in Hazel that we can import from there in case the message is changed in Hazel.

Optional as we forked Hazel.

miniduikboot (approved):

src/Impostor.Server/Net/Client.cs
js6pak: The real solution would be adding an `IsInternal` or `IsRemote` boolean property to `DisconnectedEventArgs`, but I don't think it's worth the effort right now

PR #649, opened 2024-09-07
Fix vanilla venting as Phantom cause a ban

NikoCat233:


miniduikboot (approved):
Hmm, I can see how this one slipped through the cracks. Thanks for the fix

PR #651, opened 2024-09-23
Let server serialize certain root messages instead of copying from host

NikoCat233:
Root messages like StartGame, EndGame and AlterGame should be serialized by server instead of copying from host's packets.
I believe no mods will write their stuffs in these root message,
 and server serializing these root messages is better for anticheat purpose.

### DIFF
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -198,7 +198,7 @@ namespace Impostor.Server.Net
                         return;
                     }
 
-                    await Player!.Game.HandleStartGame(reader);
+                    await Player!.Game.HandleStartGame();
                     break;
                 }
 
@@ -269,7 +269,7 @@ namespace Impostor.Server.Net
                         reader,
                         out var gameOverReason);
 
-                    await Player!.Game.HandleEndGame(reader, gameOverReason);
+                    await Player!.Game.HandleEndGame(gameOverReason);
                     break;
                 }
 
@@ -290,7 +290,7 @@ namespace Impostor.Server.Net
                         return;
                     }
 
-                    await Player!.Game.HandleAlterGame(reader, Player, value);
+                    await Player!.Game.HandleAlterGame(Player, gameTag, value);
                     break;
                 }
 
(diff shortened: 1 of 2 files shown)


NikoCat233:
image
What should I do then

NikoCat233:
Guess I need to make S2C messages in Impostor.Api. Finishing this later.

miniduikboot:
> I believe no mods will write their stuffs in these root message,

What mods do isn't really relevant, there have been cases where Impostor is permitting too much and many cases where Impostor isn't supporting certain things. What do the official servers do?

NikoCat233:
Lemme try

NikoCat233:
For rpc end game they obviously override it.

miniduikboot:
EndGame gets overriddenbecause of the cosmicube stuff, so that makes sense.

For StartGame/AlterGame however I can see the usecase, so I think we should probably check instead if the buffer contains a trailer and flag it is as a game extension in all cases when  a trailer is found. What do you think?

NikoCat233:
So we check by reading the reader with C2S deSerialize and checks bytes remaining?

miniduikboot:
yes

NikoCat233:
I still think these calls should remain vanilla way tbh

NikoCat233:
Still think copying root messages from host is risky.
Innersloth doing **** doesnt mean we have to completely follow it.

src/Impostor.Api/Net/Messages/S2C/Message02StartGameS2C.cs
```diff
@@ -0,0 +1,20 @@
...
+using Impostor.Api.Games;
+
+namespace Impostor.Api.Net.Messages.S2C
+{
+    public static class Message02StartGameS2C
+    {
+        public static void Serialize(IMessageWriter writer, GameCode gameCode)
+        {
+            writer.StartMessage(MessageFlags.StartGame);
+            gameCode.Serialize(writer);
```
miniduikboot, line 11: MAJOR: I don't see the client reading out the game mode, although it does serialize it in InnerNetClient.SendStartGame

I suggest moving this code to the deserializing part of StartGameC2S and checking if the game code matches the currently connected game

miniduikboot (changes requested):

PR #653, opened 2024-09-28
Bugfix/fix outfit RPCs

danbudworthmead:
### Description

Implemented the sequence IDs. All set calls must have a greater ID than the previous otherwise the game client disregards the rpc call.


src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
```diff
@@ -215,5 +215,44 @@ private void SetTasks(ReadOnlyMemory<byte> taskTypeIds)
...
+                    ++defaultOutfit.NamePlateSequenceId;
+                    nextRpcSequenceId = defaultOutfit.NamePlateSequenceId;
+                    break;
+                default:
+                    _logger.LogError(
+                        "{NextRpcSequenceIdName} was called for an RPC it doesn\'t know how to handle: {Rpc}",
+                        nameof(GetNextRpcSequenceId),
+                        rpc);
+                    break;
+            }
```
miniduikboot, line 254: Please fix the warning:

> Closing brace should be followed by blank line

src/Impostor.Api/Innersloth/Customization/PlayerOutfit.cs
```diff
@@ -59,15 +59,15 @@ public string NamePlateId
             internal set => SetField(ref _namePlateId, value);
         }
 
-        private byte HatSequenceId { get; set; } = 0;
+        public byte HatSequenceId { get; set; } = 0;
```
miniduikboot, line 62: Is there a good reason to expose this to the public API? Can we keep this internal to Impostor?

Users shouldn't care about the sequence ID because Impostor should send the next sequence ID

src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
```diff
@@ -215,5 +215,44 @@ private void SetTasks(ReadOnlyMemory<byte> taskTypeIds)
...
+        {
+            var defaultOutfit = this.Outfits[PlayerOutfitType.Default];
+
+            byte nextRpcSequenceId = 0;
+
+            // ReSharper disable once SwitchStatementHandlesSomeKnownEnumValuesWithDefault
+            switch (rpc)
+            {
+                case RpcCalls.SetHatStr:
+                    ++defaultOutfit.HatSequenceId;
```
miniduikboot, line 229: Is a +1 sufficient or should we add, say, +5 to make sure our write wins from a concurrent write from the client?

miniduikboot (changes requested):
I don't see the usecase in exposing the sequence id's, so Arkatme, please explain or make them internal

src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
danbudworthmead: > Is a +1 sufficient or should we add, say, +5 to make sure our write wins from a concurrent write from the client?

I've been running this PR all day on skeld.net and not occured any issues so far.

danbudworthmead:
@miniduikboot I have a usecase for making those sequence IDs public

https://github.com/danbudworthmead/net.skeld.plugins/pull/3/files

src/Impostor.Server/Net/Inner/Objects/InnerPlayerInfo.cs
miniduikboot: That doesn't answer my question. Players and the server may be racing to set a hat: if a player is setting a hat at the same time the server is setting a hat, who should win? If both of them send a hat at the same time, I think the server and the player will be desynced. Also applies to visors etc

Example from the Among Us source code: when RpcSnapTo is called, it adds 2 to the sequence id to make sure it wins from a player movement.

miniduikboot (approved):
Quickly validated the changes logically, works fine. Changes seem fine as well

PR #660, opened 2024-12-20
Add 207 XboxDeclareXuid | Correctly handle unknown gamedata tag

NikoCat233:
Added 207 XboxDeclareXuid, which previously cause a warning log

Upon receiving unhandled gamedata tag, we first check whether it is defined in vanilla protocol.
If not, we check protocol extension anticheat, reject the packet if anticheat flags, or log warning and continue broadcasting.

Change warning message to Debug and cheat detection to avoid unnecessary logging

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Data.cs
+++ b/src/Impostor.Server/Net/State/Game.Data.cs
@@ -315,7 +315,18 @@ namespace Impostor.Server.Net.State
 
                     default:
                     {
-                        _logger.LogWarning("Bad GameData tag {0}", reader.Tag);
+                        if (Enum.IsDefined(typeof(GameDataTag), reader.Tag))
+                        {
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/State/Game.Data.cs
```diff
@@ -315,7 +315,18 @@ public async ValueTask<bool> HandleGameDataAsync(IMessageReader parent, ClientPl
 
                     default:
                     {
-                        _logger.LogWarning("Bad GameData tag {0}", reader.Tag);
+                        if (Enum.IsDefined(typeof(GameDataTag), reader.Tag))
```
miniduikboot, line 318: GameDataTag is not an enum, does this work?

NikoCat233: This fix is dumb, but i dont know better ways to do this

NikoCat233:
This pr is not related to rpcflag

PR #663, opened 2024-12-27
Add diagnostic http message

miniduikboot:
### Description

Many people are struggling with setting up the HTTP server, so add a diagnostic message to allow them to test if their HTTP server is working. It's a plain HTTP page, which should suffice for now

### DIFF
--- a/src/Impostor.Server/Http/HelloController.cs
+++ b/src/Impostor.Server/Http/HelloController.cs
@@ -1,4 +1,5 @@
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
 
 namespace Impostor.Server.Http;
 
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Http/HelloController.cs
```diff
@@ -0,0 +1,32 @@
...
+    public IActionResult GetHello()
+    {
+        if (!_shownHello)
+        {
+            _shownHello = true;
+            _logger.LogInformation("Impostor's Http server is reachable (this message is only printed once per start)");
+        }
+
+        return Ok(@"Impostor is running, please configure your Among Us to connect to a game
+To generate a region file, go to https://impostor.github.io/Impostor");
```
js6pak, line 30: nit:
```suggestion
        return Ok(
            """
            Impostor is running, please configure your Among Us to connect to a game
            To generate a region file, go to https://impostor.github.io/Impostor
            """
        );
```

js6pak (approved):

Issue #667, opened 2025-01-29
Host's SendLateRejection does not correctly kick target player


NikoCat233:
Host sends MessageFlags.RemovePlayer to kick players that doesnt load shipstatus and get a ClientTimeOut Kick
MessageFlags.RemovePlayer is then handled by

> public async ValueTask HandleRemovePlayer(int playerId, DisconnectReason reason)

which only send RemovePlayerS2C to players except the target that was kicked
and target player should be kicked with MessageFlags.KickPlayer so it would disconnect itself

This is how 

> // Player can refuse to be kicked and keep the connection open, check for this.

is triggered. We only sent remove player to other players, but forget the kickplayer to target player, so it wont disconnect itself

NikoCat233:
1.kick player should only be sent to target player
2. to let the player close its connection, we need to send kickplayer in handleplayerremove

NikoCat233:
![Image](https://github.com/user-attachments/assets/fb705907-c447-4416-ae63-26213121ea0b)

NikoCat233:
This appears to be a very complex issue. HandleRemovePlayer is also called by Hazel Connection if client self disconnect, or called by anticheat. Both can't be correctly handled by current codes

PR #668, opened 2025-01-29
Fix player not being correctly disconnected

NikoCat233:
1. Add missing DisconnectReason to Message11KickPlayerS2C
2. Only send kick player message to target player


NikoCat233:
As mentioned in #667 
I dont know how to properly handle removeplayer from hazel connection, from server anticheat, or from host messages.
In different cases target client has already disconnected, or still connected & waiting for a kick.

NikoCat233:
Not sure how is innersloth officials doing this

PR #670, opened 2025-02-08
Player api extensions

ChipmunkMC:
This pull request simply adds a `SetRoleAsync` method to `IInnerPlayerControl`. It does not handle non-overridable roles yet, so I have marked this as a draft (I will try to find a way to bypass those!)

### DIFF
--- a/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
@@ -71,6 +71,22 @@ namespace Impostor.Api.Net.Inner.Objects
         ValueTask SetSkinAsync(string skinId);
 
         /// <summary>
+        ///     Sets the visor of the current <see cref="IInnerPlayerControl" />.
+        ///     Visible to all players.
+        /// </summary>
+        /// <param name="visorId">A visor for the player.</param>
+        /// <returns>Task that must be awaited.</returns>
+        ValueTask SetVisorAsync(string visorId);
+
+        /// <summary>
+        ///     Sets the nameplate of the current <see cref="IInnerPlayerControl" />.
+        ///     Visible to all players.
+        /// </summary>
+        /// <param name="nameplateId">A nameplate for the player.</param>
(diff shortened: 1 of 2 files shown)


ChipmunkMC:
Perhaps I shouldn't have re-used the branch from the previous PR... Feel free to re-create my commits if needed.

miniduikboot:
I'll look at it later, thanks for the PR

> It does not handle non-overridable roles yet,

iirc Innersloth added them to the game by accident, so they may go poof in the next major game version. So don't stress about them rn

ChipmunkMC:
I have found a bug: players newly joining lobbies to not appear to be aware of other players' roles, although it is aware of whether or not players are dead.
Also, the current code sets `RoleWhenAlive` whenever a new role is assigned to an alive player; I'm not sure if this is desired behavior.
Also, while this is not related to this PR, player roles default to `null`.

ChipmunkMC:
I found the cause of the desync, and it is a client bug: when deserializing `NetworkedPlayerInfo`, AU only sets the `RoleType` field, and not the `Role` field. I wrote a small mod to debug that (in the image the `RoleType` value is displayed).
image
If anyone can easily contact someone from Innersloth about this bug consider doing so; otherwise I might try to do so myself.

Also, I checked and the official client seems to set `RoleWhenAlive` when a non-dead role is assigned too. The default role for players is the crewmate role (as was found w/ my mod), so I will probably make Impostor default to this role instead of `null`. Also, a function for checking if a role is a ghost role should be added, probably.

NikoCat233:
its not a bug since it wont happen in normal games

ChipmunkMC:
> obviously its not innersloth's fault you are doing modification far beyond normal game behaviour
> 
> i think it is done intentionally that role behaviour will only spawn upon setrole, and in normal games roles are always set after everyone has joined and is ready.

This would make sense if the game didn't attempt to deserialize roles from `NetworkedPlayerInfo` instances in the first place. It even uses `Role.Role` instead of `RoleType` in the `Serialize` method. Because of this, I believe this was an oversight caused by role de/serialization not being used in normal game behavior, and not a design choice at all.

PR #673, opened 2025-02-21
Introduce handle game data inner to impostor

NikoCat233:
This is a example pr that handles game data packets like innersloth's client code logic.

In old code logic, when an exception occurs, the remaining messages are untouched and the whole root message is dropped.

In this new code, when an exception occured in single gamedata call, only the bad message is removed from parent message.

Similar codes have been staging on Niko Regions for a month, making it extremely stable.


NikoCat233:
This is just an example pr. I do not guarantee it to work 100%.
On my private fork there are many changes associated with codes in this pr to make the whole stuff work.

Probably if all the child messages in a parent message are removed, we should return false to drop it. I don't know how to check whether there is vaild message in a parent message so I don't write codes regarding this.

PR #675, opened 2025-02-27
Add Murder/Protect/Exile API without checks

### LINKED ISSUE #671: Unnecessary checks in InnerPlayerControl's API methods
# Api Suggestion

## Suggestion

Checks in some player API methods, such as `MurderPlayerAsync`, perform checks for certain conditions, and throw `ImpostorProtocolException`s if said checks are not met. However, Most of not all of these are not needed. For example, the Among Us client seems to accept all murders, even if said murders are performed by crewmates.

## Use case

On my personal server, I have a plugin that adds various commands (such as `/murder`) for the sake of allowing for players to easily manipulate the state of the game. Although there are likely few other usecases that exist, one I can think of is allowing ghosts to kill under certain conditions to implement some special role.

miniduikboot:
### Description

Force(.*)Async will perform the same function as \1Async but without throwing exceptions. The old API is maintained for backwards compatibility reasons.

Murder checks are now shared with the main function that handles CheckMurder, as ProtectPlayerAsync and ExileAsync aren't used in the code I didn't make a similar change there. Checks were also synced between both methods

### Closes issues

- closes #671

### DIFF
--- a/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
+++ b/src/Impostor.Api/Net/Inner/Objects/IInnerPlayerControl.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using Impostor.Api.Innersloth;
 using Impostor.Api.Innersloth.Customization;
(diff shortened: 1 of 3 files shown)


NikoCat233 (approved):

NikoCat233:
no matter what, we cant avoid throwing a exception lmao

ChipmunkMC (approved):

miniduikboot:
> no matter what, we cant avoid throwing a exception lmao

Discussed this remark in DM's. This PR shouldn't throw any exceptions from Impostor.Server code anymore, only inside API calls

miniduikboot:
Merged via #676

PR #676, opened 2025-02-27
Remove allow null from PlayerControl.PlayerInfo

miniduikboot:
### Description

For some network objects, like PlayerControl, we expect that there is an
PlayerInfo object associated to them. If there isn't, throw an exception
in this AC category.

It is possible to spawn invalid network objects without an attached
PlayerInfo, and Impostor will throw exceptions everywhere if you happen
to spawn one. Officials do allow spawning objects like this, so allow it
in Impostor for compatibility reasons

NOTE: depends on #675 as I needed to modify these brand new API's as well

### Closes issues

- makes it possible to spawn "CustomNetObjects" in mods like EHR


miniduikboot:
This makes a minor changes to the API:
- IInnerPlayerControl.PlayerInfo is now nullable. It always was, this is now reflected in the type so that plugins can also check this as well

NikoCat233 (approved):

src/Impostor.Api/Config/AntiCheatConfig.cs
```diff
@@ -24,6 +24,8 @@ public class AntiCheatConfig
 
         public bool EnableTargetChecks { get; set; } = true;
 
+        public bool EnableInvalidObjectChecks { get; set; } = true;
```
miniduikboot, line 27: MINOR: reorder to maintain alphabetic order

src/Impostor.Server/Net/Client.cs
```diff
@@ -73,6 +73,7 @@ bool LogUnknownCategory(CheatCategory category)
                 CheatCategory.Ownership => _antiCheatConfig.EnableOwnershipChecks,
                 CheatCategory.Role => _antiCheatConfig.EnableRoleChecks,
                 CheatCategory.Target => _antiCheatConfig.EnableTargetChecks,
+                CheatCategory.InvalidObject => _antiCheatConfig.EnableInvalidObjectChecks,
```
miniduikboot, line 76: MINOR: reorder to maintain alphabetic order

src/Impostor.Api/Net/Messages/S2C/Message01JoinGameS2C.cs
```diff
@@ -18,7 +18,7 @@ public static void SerializeJoin(IMessageWriter writer, bool clear, int gameCode
             writer.Write(hostId);
             writer.Write(player.Client.Name);
             player.Client.PlatformSpecificData.Serialize(writer);
-            writer.WritePacked(player.Character?.PlayerInfo.PlayerLevel ?? 1);
+            writer.WritePacked(player.Character?.PlayerInfo?.PlayerLevel ?? 1);
```
miniduikboot, line 21: OUT OF SCOPE: above lines are duplicated in Msg07JoinedGameS2C. Refactor to move to a common location?

src/Impostor.Server/Net/Inner/InnerNetObject.Anticheat.cs
```diff
@@ -73,9 +73,16 @@ protected async ValueTask<bool> ValidateCmd(CheatContext context, IClientPlayer
             return true;
         }
 
-        protected async ValueTask<bool> ValidateImpostor(CheatContext context, IClientPlayer sender, InnerPlayerInfo playerInfo, bool value = true)
+        protected async ValueTask<bool> ValidateImpostor(CheatContext context, IClientPlayer sender, InnerPlayerInfo? playerInfo, bool value = true)
         {
-            if (playerInfo.IsImpostor != value)
+            if (playerInfo == null)
+            {
+                if (await sender.Client.ReportCheatAsync(context, CheatCategory.InvalidObject, "Couldn't check if Impostor, playerInfo not set"))
```
miniduikboot, line 80: MINOR: if the invalid objects category is off but role checks is not, people are able to murder with an invalid object. Either:
1. Check both categories
2. Check Role here instead of InvObj
3. Severely warn against turning off InvObj and accept cases like this (aka documentation update)

Idem Validate{CanVent,Role}

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.Api.cs
```diff
@@ -17,10 +17,15 @@ internal partial class InnerPlayerControl : IInnerPlayerControl
...
         IInnerCustomNetworkTransform IInnerPlayerControl.NetworkTransform => NetworkTransform;
 
-        IInnerPlayerInfo IInnerPlayerControl.PlayerInfo => PlayerInfo;
+        IInnerPlayerInfo? IInnerPlayerControl.PlayerInfo => PlayerInfo;
 
         public async ValueTask SetNameAsync(string name)
         {
+            if (PlayerInfo == null)
+            {
+                throw new ImpostorProtocolException("Cannot set name, PlayerInfo is null");
```
miniduikboot, line 26: MINOR: Add to API docs

Idem for other cosmetic api's

miniduikboot (commented):
PR needs minor changes, doesn't block a production test

src/Impostor.Server/Net/Inner/InnerNetObject.Anticheat.cs
miniduikboot: Documentation update was chosen

PR #677, opened 2025-03-02
Implement an API for starting the game

ChipmunkMC:
### Description

This PR adds an API method that starts the game.

I have tested it with my own copy of Among Us, and it works as expected.
image

### Closes issues

- closes #

PR #678, opened 2025-03-03
Make game ended events cancellable

ChipmunkMC:
This PR makes it possible to cancel game ended events, effectively allowing the server to control the conditions for the game ending. This pairs nicely with #677 as it can be used to prevent the game from ending after being forcefully started with less than 4 players.


ChipmunkMC:
I have uploaded a demonstration [here](https://files.chipmunk.land/selif/vulfpn6b.mkv) in case it is of interest. (GitHub did not let me attach the video and I couldn't be bothered to convert it, so I used my own service instead)

ChipmunkMC:
This causes meetings to behave weirdly by the way, if you want I could look into workarounds, or we could let plugins handle it.

PR #679, opened 2025-03-06
Fix spawntimeout not run for rejoining players

NikoCat233:
1. force clearing client character in case hosts in +25 does not despawn it
2. use HandleJoinGameNew to pull in limbo players. This produces clear logs and correctly triggers spawn time out.

### DIFF
--- a/src/Impostor.Server/Net/State/Game.State.cs
+++ b/src/Impostor.Server/Net/State/Game.State.cs
@@ -133,16 +133,10 @@ namespace Impostor.Server.Net.State
 
         private async ValueTask CheckLimboPlayers()
         {
-            using var message = MessageWriter.Get(MessageType.Reliable);
-
-            foreach (var (_, player) in _players.Where(x => x.Value.Limbo == LimboStates.WaitingForHost))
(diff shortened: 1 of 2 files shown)


src/Impostor.Server/Net/State/Game.State.cs
```diff
@@ -133,16 +133,10 @@ private async ValueTask MigrateHost()
...
+            var limboPlayers = _players.Where(x => x.Value.Limbo == LimboStates.WaitingForHost);
+            foreach (var (_, player) in limboPlayers)
             {
-                WriteJoinedGameMessage(message, true, player);
-                WriteAlterGameMessage(message, false, IsPublic);
-
-                player.Limbo = LimboStates.NotLimbo;
-
-                await SendToAsync(message, player.Client.Id);
+                await HandleJoinGameNext(player, false);
```
HayashiUme: This causes non-host players to remain in the Limbo state indefinitely when they are added to the limboPlayers list before the host returns to the lobby

NikoCat233: Correct, I never noticed this since in my fork I used HandleJoinGameNew

NikoCat233:
chatgpt reviewed and think this fine

PR #680, opened 2025-03-08
Add a proper API for IInnerPlayerPhysics

ChipmunkMC:
This PR adds an API for IInnerPlayerPhysics, which can be used to interact with vents, ladders, and pets.


Pietrodjaowjao:
What if you also add a modified version of EnterVentAsync that teleports the player to it before so it doesnt need to walk to?

ChipmunkMC:
> What if you also add a modified version of EnterVentAsync that teleports the player to it before so it doesnt need to walk to?

I thought of it, that might be useful. I might do that later

ChipmunkMC:
Also, I found out that the `EnterVentAsync` method does not make the navigation arrows appear, and bypasses the hide and seek vent timer. If anyone wants I can add it to the docs, or if you are a maintainer feel free to edit them yourself if you want.

PR #686, opened 2025-04-24
Add client network version check API

yukieiji:
### Description

Added API to check client version in HelloController
This makes it possible to check the compatibility of the Impostor server with the AmongUs running on mods, etc., before connecting, and to hide the target server, etc., if necessary.

### Closes issues

- closes NONE

### DIFF
--- a/src/Impostor.Server/Http/HelloController.cs
+++ b/src/Impostor.Server/Http/HelloController.cs
@@ -1,5 +1,11 @@
+using System.Reflection;
+using System.Text.Json.Serialization;
+using Impostor.Api.Config;
+using Impostor.Api.Innersloth;
+using Impostor.Api.Net.Manager;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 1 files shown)


miniduikboot (changes requested):
I'm not sure I see the point of putting this into the core of Impostor. This code is not used by vanilla Among Us, so mods will need to integrate this themself. We generally don't do extensions like this, and prefer to do this in a plugin (see Reactor.Impostor)

I see two ways forwards:
- We could refuse to hand out a token if the version is not compatible, but I'm not sure if the UX will be better compared to the current method
- This could be implemented as a plugin (see Boot.Metrics as an example of how to piggyback on Impostor's ASP.NET Core server)

yukieiji:
Yes, it doesn't exist in vanilla AmongUs and can be implemented with a plugin.
However, by implementing the API in the core instead of in a plugin, it is possible to deploy services that use the API, and there is no need to think about divergence of processing in the mod depending on the presence or absence of plugins.

yukieiji:
Thanks for the advice.

I have already created [the plugin](https://github.com/yukieiji/CompatAPI) itself and will refer to it for future updates

PR #690, opened 2025-05-20
Add 16.0.5 support

miniduikboot:
### Description

Crossplay with the previous version is not supported, so it gets its own compatibility group.

Further analysis on the changes is still needed

### Closes issues

- https://www.innersloth.com/may-update-a-smol-treat/


NikoCat233:
No base game protocol difference

miniduikboot:
There is a lot of noise in the ILSpy diff, but it seems like there are no server side changes necessary

PR #691, opened 2025-05-22
Don't clear RoleOptions on update

miniduikboot:
### Description

Among Us always adds a constant amount of roles to this dict, so it doesn't make sense to clear this dictionary every time.

Furthermore, previously it was possible to trigger a ConcurrentModificationException when running Deserialize in parallel with Serialize. This should no longer happen, as new keys are no longer added.


NikoCat233 (approved):

miniduikboot:
Ran it in production for half a week, seems stable

PR #698, opened 2025-08-07
Add Impostor Status Panel

FangkuaiYa:
4b0f1bc59ad9a5574b30f5b2f3a2c822
47121a0397c82c386097bc5c5ceac553
4082d58dca4209da9b0ecf3136901a41

### DIFF
--- a/src/Impostor.Server/Http/HelloController.cs
+++ b/src/Impostor.Server/Http/HelloController.cs
@@ -1,20 +1,26 @@
+using System;
+using System.ComponentModel;
+using System.Linq;
+using Impostor.Api.Games;
+using Impostor.Api.Games.Managers;
+using Impostor.Api.Innersloth;
+using Impostor.Server.Utils;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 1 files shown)


FangkuaiYa:
:(

miniduikboot:
Thanks for your contribution. While I see the value of this, I don't think it's something that should be enabled on every server by default. Generally we prefer to keep things that are not core to Impostor (like Reactor compatibility, metrics or HTTP matchmaking - as it was initially) outside of Impostor itself and in a plugin. HelloController is pushing the limits of this policy already, but it's proven very useful as a checkpoint when trying to debug an Impostor setup.

Therefore I'd recommend you to create this as a plugin.

If you'd want to implement this as a plugin, you could create an ASP.NET Core controller class, make your plugin startup class inherit from IPluginHttpStartup and add the route in the ConfigureWebApplication hook.

miniduikboot:
> > Also some info should not be shown publicly
> 
> Generally speaking, you should avoid displaying the server's IP address directly. 通常来说，你应该将服务器的IP地址脱敏处理。

I'm not concerned about the IP address, this can be retrieved in various easy ways. Listing all (private!) games on the server is much more problematic

PR #699, opened 2025-08-10
Add IsCancelled to task completion event; gate CompleteTask RPC when …

ABraik-bit:
…cancelled

### Closes issues

- closes #

### DIFF
--- a/src/Impostor.Api/Events/Game/Player/IPlayerCompletedTaskEvent.cs
+++ b/src/Impostor.Api/Events/Game/Player/IPlayerCompletedTaskEvent.cs
@@ -2,7 +2,7 @@
 
 namespace Impostor.Api.Events.Player
 {
-    public interface IPlayerCompletedTaskEvent : IPlayerEvent
+    public interface IPlayerCompletedTaskEvent : IPlayerEvent, Impostor.Api.Events.IEventCancelable
     {
         ITaskInfo Task { get; }
     }
--- a/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
@@ -615,14 +615,14 @@ namespace Impostor.Server.Net.Inner.Objects
             ProtectedBy = guardianAngel;
         }
 
-        private async ValueTask HandleCompleteTask(ClientPlayer sender, uint taskId)
+        private async ValueTask<bool> HandleCompleteTask(ClientPlayer sender, uint taskId)
         {
             TaskInfo? task = null;
             if (PlayerInfo == null)
             {
                 if (await sender.Client.ReportCheatAsync(RpcCalls.CompleteTask, CheatCategory.InvalidObject, "PlayerControl doesn't have PlayerInfo"))
                 {
-                    return;
+                    return false;
                 }
             }
             else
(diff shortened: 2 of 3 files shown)


src/Impostor.Api/Events/Game/Player/IPlayerCompletedTaskEvent.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Impostor.Api.Events.Player
 {
-    public interface IPlayerCompletedTaskEvent : IPlayerEvent
+    public interface IPlayerCompletedTaskEvent : IPlayerEvent, Impostor.Api.Events.IEventCancelable
```
miniduikboot, line 5: Please just add a `using` directive instead of qualifying it like this

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -615,14 +615,14 @@ internal void Protect(InnerPlayerControl guardianAngel)
             ProtectedBy = guardianAngel;
         }
 
-        private async ValueTask HandleCompleteTask(ClientPlayer sender, uint taskId)
```
miniduikboot, line 618: Return value of this function is ignored in HandleRpcAsync

miniduikboot (changes requested):
I agree in principle with the PR, 2 minor remarks

miniduikboot:
Hmm, that's true, thanks for bringing it up. Cancelling this event doesn't update the internal state: the client will think it has completed the task, the server also thinks the task is complete as task.Complete is set to true... The only group that doesn't know that a task was completed are the other players in the match.

From what I understood from Aiden his usecase is to delay broadcasting task completions.

Maybe we should define cancelable better:
- Does this stop the packet from being broadcasted further (currently: yes)
- Does this stop the event from occurring at all for anyone in the match? (currently: in this case no, for other events: mostly yes - GameCreation stops creating the game at all, PlayerCheckMurder and PlayerChat will not broadcast it, but the sender will still see the message as sent)

I see 4 states:
1. The event goes through completely and nothing gets cancelled
2. The event is processed by the server and the sender, but not by the other clients (desyncs rest of match - currently this PR)
3. The event is processed by the sender, but not by the server or other clients (desyncs sender - currently PlayerCheckMurder and PlayerChat? We don't maintain chat state on the server)
4. The event is processed by nobody (currently GameCreation)

Should Impostor API's exist that desync clients from the server, and if yes, because this is not the first usecase asking for this, how do we do this in a nice way? Should there be a separate interface for "Hold my beer, my plugin will handle it"-style of cancellations?

miniduikboot:
So here is how I suggest to move forward:
- Event cancellation should stop Impostor from handling the request further: so don't broadcast it and - as we're currently already doing - stop handling events for it.
- I'm split on keeping everything using IEventCancellable or whether there should be a separate interface for cancelling broadcasts. If we're doing the latter, we should consider move over PlayerChat and PlayerCheckMurder (or make that event uncancellable, as that's probably not what people want)
- If internal state can be updated, the new value should be settable in the event (example: game codes in GameCreation)
- We document the impact of event cancellation for each cancellable event. If it desyncs players like this event, we add this in the documentation, so plugin devs know the impact and what they'd need to do to fix it.
- Serializing PlayerInfo's on an event cancellation like this is a bit magic, I don't think that's a good path to go to, as it may fight with people's plugins. Exposing an easy method to fix it would be a good option

PR #704, opened 2025-09-24
Fix HandshakeC2S

NikoCat233:
## Title
Remove unused Version compare values and redundant reader reads

## Description
Based on analysis of assemblies from 2024.6.18 and later versions, the following reader operations are no longer meaningful and have never worked correctly. They can be safely removed:

- `reader.ReadInt32(); // crossplayFlags, not used yet`
- `reader.ReadByte(); // purpose unknown, seems hardcoded to 0`

Now, the message only needs to be read up to `platformData`. Any remaining data in the message can be safely ignored.

## Main Changes
- Removed the above two unnecessary reader read operations.
- Simplified version compare and message parsing logic for easier future maintenance.

## Reason for Change
- These values are no longer used in new assemblies, and reading them may cause confusion or compatibility issues.
- The code is now cleaner and more maintainable.

## Compatibility Notes
- This change only affects message parsing logic and does not impact core functionality or compatibility with newer versions.

## Related Issues
None

## Checklist
- [x] Main flow tested and works correctly
- [x] Compiles successfully
- [x] Other related modules updated if required

Please review and provide feedback. Thank you!

### DIFF
--- a/src/Impostor.Api/Net/Messages/C2S/HandshakeC2S.cs
+++ b/src/Impostor.Api/Net/Messages/C2S/HandshakeC2S.cs
@@ -25,7 +25,7 @@ namespace Impostor.Api.Net.Messages.C2S
             using var platformReader = reader.ReadMessage();
             platformSpecificData = new PlatformSpecificData(platformReader);
 
-            if (clientVersion.Normalize() == Version.V3)
+            if (clientVersion.Normalize() <= Version.V3)
             {
                 // Crossplay flag was removed in 2021.12.14, friendcode was added instead
                 reader.ReadInt32();


Pietrodjaowjao:
That one hardcoded to 0 had a meaning in 2019, when Mira HQ map was a paid map, 0 would mean that the host doesnt have mira hq and 1 would mean that he did

src/Impostor.Api/Net/Messages/C2S/HandshakeC2S.cs
```diff
@@ -9,48 +9,15 @@ public static void Deserialize(IMessageReader reader, out GameVersion clientVers
...
-        private static class Version
-        {
-            public static readonly GameVersion V1 = new(2021, 4, 25);
-
-            public static readonly GameVersion V2 = new(2021, 6, 30);
-
-            public static readonly GameVersion V3 = new(2021, 11, 9);
-
-            public static readonly GameVersion V4 = new(2021, 12, 14);
+            // Skip the rest of the message (empty friendcode and placeholder uint 0)
```
miniduikboot, line 20: This is not acceptable, the reader needs to advance to the end of the message, otherwise plugins like Reactor.Impostor that parse handshake trailers will fail

src/Impostor.Api/Net/Messages/C2S/HandshakeC2S.cs
```diff
@@ -9,48 +9,15 @@ public static void Deserialize(IMessageReader reader, out GameVersion clientVers
             clientVersion = reader.ReadGameVersion();
             name = reader.ReadString();
 
-            if (clientVersion >= Version.V1)
-            {
-                reader.ReadUInt32(); // lastNonceReceived, always 0 since 2021.11.9
-            }
+            reader.ReadUInt32(); // lastNonceReceived, always 0 since 2021.11.9
```
miniduikboot, line 12: The reason the if-statements were in place is to help old clients realize that they need to update, and disconnect them with the proper message, instead of with a server exception

miniduikboot (changes requested):
I don't see a reason to merge this PR. The choices made in https://github.com/Impostor/Impostor/commit/3065e8df48e340e960199bfc557618233dd352f5 were intentional and I see no reason to change this code.

src/Impostor.Api/Net/Messages/C2S/HandshakeC2S.cs
NikoCat233: Before Version.V3, handshake is serialized with binwriter, which definately cannot be read now.

PR #719, opened 2026-03-23

GoobisMoobis:
### Description

If this produces unforeseen errors I can refine my change


NikoCat233:
can you open a new pr with correct commit message?
like you should say "fix a race condition in ..." instead of "Update Game.Data.cs"

NikoCat233 (changes requested):
fix makes sense. But need to correct commit message

GoobisMoobis:
> can you open a new pr with correct commit message?
> 
> like you should say "fix a race condition in ..." instead of "Update Game.Data.cs"

Once I get home tonight I will

PR #720, opened 2026-03-23
Goobismoobis race

GoobisMoobis:
### Description

If this produces unforeseen errors I can refine my change.

I also fixed the commit message to be more descriptive than my last pull request


GoobisMoobis:
DAMN IT I FORGOT TO CHANGE THE NAME OF THE PULL

PR #721, opened 2026-03-23

GoobisMoobis:
### Description

If this produces unforeseen errors I can refine my change.

I also fixed the commit message to be more descriptive than [my last pull request](https://github.com/Impostor/Impostor/pull/719)

I also made the title of this pull request better than I did for [this one](https://github.com/Impostor/Impostor/pull/720)

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Data.cs
+++ b/src/Impostor.Server/Net/State/Game.Data.cs
@@ -468,7 +468,7 @@ namespace Impostor.Server.Net.State
 
         private async ValueTask SyncServerObjectsAsync(ClientPlayer sender)
         {
-            foreach (var obj in _allObjectsFast.Values)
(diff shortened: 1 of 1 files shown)


NikoCat233:
Create a new branch in your repo, manually commit amd open a new pr

NikoCat233 (changes requested):
didnt fix the commit message

GoobisMoobis:
> didnt fix the commit message

Oh, I thought I did fix it by reverting the change and re-committing with a different message. For some reason GitHub doesn't put that commit at the top of my pull, but because it was the most recent that's the one that'll show on the name of the file.

GoobisMoobis:
> Create a new branch in your repo, manually commit amd open a new pr

I'll do this rn

PR #722, opened 2026-03-24

GoobisMoobis:
I'll attach my old pull request description here:

> ### Description
> Fixes an error when 2 players join at the same time by using ToList(), which makes it so the the loop iterates an independent list that can't be affected by 2 players touching the same dictionary during the awaits.
> 
> If this produces unforeseen errors I can refine my change.
> 
> I also fixed the commit message to be more descriptive than [my last pull request](https://github.com/Impostor/Impostor/pull/719)
> 
> I also made the title of this pull request better than I did for [this one](https://github.com/Impostor/Impostor/pull/720)

After reading the feedback on my most recent pull, I realized the reason you wanted me to give a detailed commit message was probably to keep the commit history clean, NOT just so that it would show as the most recent edit to the file.

However, my first patch, I had completely forgot to create a new branch before I did my first messy commit, so I ended up deleting the entire old repo. This commit comes from my new one with a much cleaner commit.

This pull should hopefully be the last of this saga. Sorry NikoCat 😬

### DIFF
--- a/src/Impostor.Server/Net/State/Game.Data.cs
+++ b/src/Impostor.Server/Net/State/Game.Data.cs
@@ -468,7 +468,7 @@ namespace Impostor.Server.Net.State
 
         private async ValueTask SyncServerObjectsAsync(ClientPlayer sender)
         {
-            foreach (var obj in _allObjectsFast.Values)
+            foreach (var obj in _allObjectsFast.Values.ToList())
             {
                 if (obj.OwnerId == ServerOwned)
                 {


miniduikboot:
Thanks, I'll start a production test to see whether it helps reducing the amount of exceptions I see

AeonLucid:
What about using a ConcurrentDictionary for `_allObjectsFast`?
Instead of copying the entire list for every `SyncServerObjectsAsync` call.

miniduikboot:
Iterating over a ConcurrentDictionary also allocates a new copy: https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L2165 so it doesn't matter if you want to get bitten by the cat or by the dog.

This fix is a bit minimal and there are likely more to follow until we close the issue at its root: multiple threads can process messages on the same room at the same time. This will likely need a larger rework, either using synchronization to prevent issues or creating a task for each game that will process each message for that room, effectively making each room single threaded.

Initial results from production are hopeful, 0 exceptions on the fixed server, 7 on the control server.

miniduikboot:
Hi GoobisMoobis, thanks for your PR. Unfortunately I've decided to deny it as it doesn't solve all the issues related to concurrency in this function. I changed it to ConcurrentDictionary and now it's stable enough. Thanks for doing the investigative work

PR #726, opened 2026-05-12
Add support for PackedGameDataTo

NikoCat233:
very very basic support for packed game data to
https://github.com/Innersloth-LLC/AmongUsModdingInformation


src/Impostor.Server/Net/Client.cs
```diff
@@ -407,6 +449,29 @@ private bool IsPacketAllowed(IMessageReader message, bool hostOnly)
...
+            var game = Player.Game;
+
+            if (message.ReadPackedInt32() != game.Code)
+            {
+                return false;
+            }
+
+            if (game.HostId == Id)
+            {
+                return true;
```
miniduikboot: Do we want to allow this for non-hosts or a non +25 host? Would be a shame if this capability was limited to just +25
+25 started as just a protocol hack to let hosts directly decide on kill outcomes and now it's getting abused for all kinds of AC exceptions and extra bells and whistles like /cmd

IMO this should be behind an AC check

src/Impostor.Server/Net/Client.cs
```diff
@@ -407,6 +449,29 @@ private bool IsPacketAllowed(IMessageReader message, bool hostOnly)
             return true;
         }
 
+        private bool IsPackedGameDataToAllowed(IMessageReader message)
```
miniduikboot: Code clone of IsPacketAllowed(message, true)

miniduikboot (commented):
2 conflicting remarks. Let me check in with the all-client mod community whether this is useful for them too.

src/Impostor.Server/Net/Client.cs
NikoCat233: I asked Innersloth about this two days ago, and they’re obviously intentionally not responding.

From tests on vanilla regions, Packed Game Data is available on both +0 and +25, and it can be sent by both hosts and clients.

It just makes overloading easier, and it’s already being abused as a rate-limit bypass, thanks to the groups that keep promoting their shit for money and fame.

I personally banned this feature on the +0 protocol in my regions and am still testing its effects in staging. I still need to see how the modding community responds.

src/Impostor.Server/Net/Client.cs
NikoCat233: This can be removed as long as we decide to allow the feature on both +0 and +25, and for both hosts and non-hosts.

src/Impostor.Server/Net/Client.cs
miniduikboot: I see the potential for abuse but I think we should also make sure that our software is compatible with officials. Should we put PackedGameDataTo behind an anticheat check instead of hard forbidding it? Then AllowCheatingHosts will allow it on +0/+25 for hosts and a new option AllowHostOnlyExtensions will allow it for all players.

miniduikboot: I've made up my mind:
- usage for +0 and +25 hosts will be gated behind the new Anticheat.AllowHostOnlyExtensions option, which has similar values to AllowCheatingHosts: Never/IfRequested/Always but defaults to IfRequested.
- usage for players will be forbidden for now as I don't see a usecase other than overloading clients/servers

src/Impostor.Server/Net/Client.cs
miniduikboot: Removed it in favor of a separate AC check

NikoCat233:
Patches are good

miniduikboot (approved):

Issue #727, opened 2026-05-21
Add messageType check in HandleRpcAsync and InnerNetObject.DeserializeAsync


NikoCat233:
Vanilla game sends most RPC calls and net object data using reliable messages. However, cheaters are abusing unreliable messages to spam these calls as an overload bypass method.

Therefore, it seems necessary to pass the message type to `HandleRpcAsync` and `DeserializeAsync`, and add a protocol extension check there.

miniduikboot:
Sure, sounds like a good idea, do you want to work on this?

PR #728, opened 2026-05-25
Add check for message type in handle rpc

NikoCat233:
very early implementation for #727

### DIFF
--- a/src/Impostor.Server/Net/Inner/InnerNetObject.Anticheat.cs
+++ b/src/Impostor.Server/Net/Inner/InnerNetObject.Anticheat.cs
@@ -73,6 +73,19 @@ namespace Impostor.Server.Net.Inner
             return true;
         }
 
+        protected async ValueTask<bool> ValidateReliable(CheatContext context, IClientPlayer sender, MessageType messageType)
+        {
+            if (messageType != MessageType.Reliable)
+            {
+                if (await sender.Client.ReportCheatAsync(context, CheatCategory.ProtocolExtension, "Client sent reliable-only game data using an unreliable message"))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
         protected async ValueTask<bool> ValidateImpostor(CheatContext context, IClientPlayer sender, InnerPlayerInfo? playerInfo, bool value = true)
         {
             if (playerInfo == null)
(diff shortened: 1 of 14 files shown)


src/Impostor.Server/Net/Inner/InnerNetObject.Anticheat.cs
```diff
@@ -73,6 +73,19 @@ protected async ValueTask<bool> ValidateCmd(CheatContext context, IClientPlayer
             return true;
         }
 
+        protected async ValueTask<bool> ValidateReliable(CheatContext context, IClientPlayer sender, MessageType messageType)
+        {
+            if (messageType != MessageType.Reliable)
+            {
+                if (await sender.Client.ReportCheatAsync(context, CheatCategory.ProtocolExtension, "Client sent reliable-only game data using an unreliable message"))
```
miniduikboot, line 80: I think we need to revisit what the ProtocolExtension means. Looking at the current uses, we use it in 5 places:
- Client sent unregistered call (Used by mods that add custom rpcs, basically everything that isn't using Reactor RPC's)
- Client interacted with nonexistent vent (For ToU roles like Miner that add extra vents)
- Client interacted with vent on unknown map (Submerged compat)
- Client sent unknown color (CheckColor, for "rainbow" mods that add additional player colors)
- Client sent unknown color (but this time for SetColor)
All of these support mods that are RequiredOnAllSides, as they extend the network protocol in harmless ways that only take effect when everyone is implementing these changes

But this PR and #732 are each adding their own checks in this category, and both are more checking for exploits than harmless extensions to the network protocol. I understand that Host-Only mods rely on these extensions because of limitations on official servers, but I don't think that justifies adding them into this category.

How do you suggest to proceed? Is the "Category" model bankrupt, and do we need something even more granular? Giving each rule their own id? Or do we throw these checks into the new "exploits" category?

src/Impostor.Server/Net/Inner/Objects/InnerPlayerControl.cs
```diff
@@ -105,8 +106,13 @@ public override async ValueTask DeserializeAsync(IClientPlayer sender, IClientPl
             PlayerId = reader.ReadByte();
         }
 
-        public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, ClientPlayer? target, RpcCalls call, IMessageReader reader)
+        public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, ClientPlayer? target, RpcCalls call, IMessageReader reader, MessageType messageType)
         {
+            if (call != RpcCalls.PlayAnimation && !await ValidateReliable(call, sender, messageType))
```
miniduikboot, line 111: This doesn't work: if this RPC is PlayAnimation, the switch block below will not be called

miniduikboot (changes requested):

miniduikboot:
I'm not opposed to these changes, but we need to discuss these flaws before we merge this

PR #729, opened 2026-06-06
Add compat versions for 17.4

miniduikboot:
### Description

Mobile has a different version than PC for reasons unknown at the moment, probably related to ads.


paulbrause:
Is it possible to use this fix with Docker?

miniduikboot:
Thanks for the remindrr, now you can

paulbrause:
Thank you! :)

PR #731, opened 2026-06-18
Added an RPC packet size anti-cheat check with config

HayashiUme:
## Why it exist
Malicious clients could send oversized RPC packets to the server, causing processing delays and eventual disconnection
of all players.

 ## How it works

When the server receives an RPC message inside a GameData packet, it checks reader.Length (total sub-message size)
against the configured LimitRate. If exceeded:
- The sender is flagged for cheating via ReportCheatAsync
- The player is kickedfrom the game

### DIFF
--- a/src/Impostor.Api/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Api/Config/AntiCheatConfig.cs
@@ -30,6 +30,8 @@ namespace Impostor.Api.Config
 
         public bool ForbidProtocolExtensions { get; set; } = true;
 
-        public int LimitRate { get; set; } = 1500;
+        public bool EnablePacketSizeChecks { get; set; } = true;
+
+        public int PacketSizeLimit { get; set; } = 1168;
     }
 }
--- a/src/Impostor.Api/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Api/Config/AntiCheatConfig.cs
@@ -29,5 +29,7 @@ namespace Impostor.Api.Config
         public bool EnableTargetChecks { get; set; } = true;
 
         public bool ForbidProtocolExtensions { get; set; } = true;
+
+        public int LimitRate { get; set; } = 1500;
     }
 }
--- a/src/Impostor.Api/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Api/Config/AntiCheatConfig.cs
@@ -32,6 +32,6 @@ namespace Impostor.Api.Config
 
         public bool EnablePacketSizeChecks { get; set; } = true;
 
-        public int PacketSizeLimit { get; set; } = 1153;
+        public int PacketSizeLimit { get; set; } = 1203;
     }
 }
--- a/src/Impostor.Api/Config/AntiCheatConfig.cs
+++ b/src/Impostor.Api/Config/AntiCheatConfig.cs
@@ -32,6 +32,6 @@ namespace Impostor.Api.Config
 
         public bool EnablePacketSizeChecks { get; set; } = true;
 
-        public int PacketSizeLimit { get; set; } = 1168;
+        public int PacketSizeLimit { get; set; } = 1153;
     }
 }
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -80,7 +80,7 @@ namespace Impostor.Server.Net
                     CheatingHostMode.Never => true,
                     _ => true,
                 },
-                CheatCategory.PacketSize => true,
+                CheatCategory.PacketSize => _antiCheatConfig.EnablePacketSizeChecks,
                 CheatCategory.Other => true,
                 _ => LogUnknownCategory(category),
             };
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -238,18 +238,6 @@ namespace Impostor.Server.Net
                         return;
                     }
 
-                    // Check outer message size against the configured limit.
-                    if (_antiCheatConfig.Enabled && reader.Length > _antiCheatConfig.LimitRate)
-                    {
-                        if (await ReportCheatAsync(
-                                new CheatContext(MessageFlags.FlagToString(flag)),
-                                CheatCategory.PacketSize,
-                                $"Message size {reader.Length} exceeds limit {_antiCheatConfig.LimitRate}"))
-                        {
-                            return;
-                        }
-                    }
-
                     var toPlayer = flag == MessageFlags.GameDataTo;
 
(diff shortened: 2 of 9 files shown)


src/Impostor.Server/Net/Client.cs
```diff
@@ -237,6 +238,18 @@ public override async ValueTask HandleMessageAsync(IMessageReader reader, Messag
                         return;
                     }
 
+                    // Check outer message size against the configured limit.
+                    if (_antiCheatConfig.Enabled && reader.Length > _antiCheatConfig.LimitRate)
```
NikoCat233: You can see that the messages are already being read from the parent message in either `Game.Data.cs` or `Client.cs`, so checking the length here is inappropriate, it's already sub message.

The correct place to perform this check is in `src\Impostor.Server\Net\Hazel\HazelConnection.cs`.

For example:

```csharp
private async ValueTask ConnectionOnDataReceived(DataReceivedEventArgs e)
{
    if (!IsConnected || Client == null)
    {
        return;
    }

    /*
    if (e.Message.Length > 1468)
    {
        if (
        await Client.ReportCheatAsync(
            new CheatContext("RootMessage"),
            CheatCategory.ForSure,
            "Received a message that is too large, length: " + e.Message.Length))
        {
            return;
        }
    }
    */
}
```

src/Impostor.Api/Config/AntiCheatConfig.cs
```diff
@@ -29,5 +29,7 @@ public class AntiCheatConfig
         public bool EnableTargetChecks { get; set; } = true;
 
         public bool ForbidProtocolExtensions { get; set; } = true;
+
+        public int LimitRate { get; set; } = 1500;
```
NikoCat233: Innersloth requires "Full MessageWriter packet size, including header must be less than or equal to 1200 bytes", so if you are checking at Impostor.Server level by reading message.Length, the proper limit should be 1168 bytes (32 bytes message header including ip header and udp header).

NikoCat233: You need a anticheat option to disable this MTU check

NikoCat233 (changes requested):

src/Impostor.Server/Net/Client.cs
HayashiUme: Done, moved the check to HazelConnection.ConnectionOnDataReceived

src/Impostor.Api/Config/AntiCheatConfig.cs
HayashiUme: Done, changed default to 1168

HayashiUme: Done, added EnablePacketSizeChecks.

src/Impostor.Server/Net/Hazel/HazelConnection.cs
```diff
@@ -58,6 +63,19 @@ private async ValueTask ConnectionOnDataReceived(DataReceivedEventArgs e)
                 return;
             }
 
+            // Check raw message size against the configured limit.
+            // Innersloth requires full packet ≤ 1200 bytes (1168 bytes payload after 32 bytes IP+UDP headers).
+            if (_antiCheatConfig.Enabled && _antiCheatConfig.EnablePacketSizeChecks && e.Message.Length > _antiCheatConfig.LimitRate)
```
NikoCat233: You have 
```CheatCategory.PacketSize => _antiCheatConfig.EnablePacketSizeChecks```
in ReportCheatAsync, _antiCheatConfig.EnablePacketSizeChecks here is unnecessary

File: src/Impostor.Server/config.json [hunk 4]
```diff
@@ -21,7 +21,9 @@
     "EnableOwnershipChecks": true,
     "EnableRoleChecks": true,
     "EnableTargetChecks": true,
-    "ForbidProtocolExtensions": true
+    "ForbidProtocolExtensions": true,
+    "EnablePacketSizeChecks": true,
+    "LimitRate": 1168
```
NikoCat233: Use a more detailed name like PacketSizeLimit

NikoCat233 (changes requested):

HayashiUme (commented):
Thanks for the additional feedback! Both points addressed:

- Removed `EnablePacketSizeChecks` from the category switch — it's already checked before `ReportCheatAsync` is called, so the switch just returns `true` now.
- Renamed `LimitRate` to `PacketSizeLimit` for clarity.

(Force-pushed to keep the commit history clean.)

NikoCat233:
Please follow the existing code style.

Since you added a new cheatCategory called PacketSize, the anti-cheat enable option should be in 'Client.cs', not where the actual checks are performed.

HayashiUme:
> Please follow the existing code style.
> 
> Since you added a new cheatCategory called PacketSize, the anti-cheat enable option should be in 'Client.cs', not where the actual checks are performed.

If I’ve understood correctly, this commit will fully adhere to the existing coding style. Thank you, Niko

src/Impostor.Api/Config/AntiCheatConfig.cs
```diff
@@ -29,5 +29,9 @@ public class AntiCheatConfig
         public bool EnableTargetChecks { get; set; } = true;
 
         public bool ForbidProtocolExtensions { get; set; } = true;
+
+        public bool EnablePacketSizeChecks { get; set; } = true;
+
+        public int PacketSizeLimit { get; set; } = 1168;
```
miniduikboot: Please add these fields to the documentation

https://github.com/Impostor/Impostor/blob/master/docs/Server-configuration.md#anticheat

src/Impostor.Api/Config/AntiCheatConfig.cs
miniduikboot: Niko, I'm not sure how you're getting to 32:
- Ethernet II is 14 + 4 bytes
- IPv4 is 20 bytes
- UDP is 8 bytes

- Reliable packets are 3 bytes (SendOption + 16 byte reliable id)
- Unreliable packets are 1 byte (SendOption)

HayashiUme:
I had a calculate, it should be 1153 bytes, right?
Docs done.

NikoCat233:
I will check the exact limit on innerslop's server

HayashiUme:
> I will check the exact limit on innersloop's server

Thanks for your hard work🫡

NikoCat233:
From my testing, the kick threshold on Innerslop's server is 1204 bytes or higher.
messageWriter.length >= 1204 will trigger an anti-cheat kick.
The exact threshold may vary under different network conditions.
They probably be checking at hazel level

PR #732, opened 2026-06-18
Add length check for voting complete

NikoCat233:
Saw the cheat here:
https://github.com/MrDiamond64/Hydra/commit/654ef982f2be176d79aa87e67c7543db39779a65#diff-114a4023cd7709d091a2dda6e6601d00fc30d4ffe80b8ca9374b62bbbded99d6R141

Thanks to HydraMenu for revealing this exploit.

Basically, if a client serializes an extremely large value for the `voterStates` length in the `RPC Voting Complete` packet, the game will pre-allocate a massive amount of memory for the array. Repeated abuse of this behavior can lead to excessive memory allocation and eventual memory exhaustion.

To prevent this, we validate the value against `GameOptions.MaxPlayers`. This avoids false positives while effectively blocking the exploit.

### DIFF
--- a/src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
+++ b/src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
@@ -132,6 +132,18 @@ namespace Impostor.Server.Net.Inner.Objects
                         return false;
                     }
 
+                    var position = reader.Position;
+                    var statesLength = reader.ReadPackedInt32();
+                    reader.Seek(position);
+
+                    if (statesLength < 0 || statesLength > Game.Options.MaxPlayers)
+                    {
+                        if (await sender.Client.ReportCheatAsync(call, CheatCategory.ProtocolExtension, $"Client sent {nameof(RpcCalls.VotingComplete)} with invalid voter state length {statesLength}, max allowed is {Game.Options.MaxPlayers}"))
+                        {
+                            return false;
+                        }
+                    }
+
                     Rpc23VotingComplete.Deserialize(reader, out var states, out var playerId, out var tie);
                     foreach (var messageReader in states)
                     {


NikoCat233:
This type of attack can also be implemented through many other paths. However, since we need to be able to deserialize `UpdateSystem` in order to add the appropriate validation checks, I won't continue the analysis here.

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
```diff
@@ -132,6 +132,18 @@ public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, Client
                         return false;
                     }
 
+                    var position = reader.Position;
+                    var statesLength = reader.ReadPackedInt32();
+                    reader.Seek(position);
+
+                    if (statesLength < 0 || statesLength > Game.Options.MaxPlayers)
+                    {
+                        if (await sender.Client.ReportCheatAsync(call, CheatCategory.ProtocolExtension, $"Client sent {nameof(RpcCalls.VotingComplete)} with invalid voter state length {statesLength}, max allowed is {Game.Options.MaxPlayers}"))
```
miniduikboot, line 141: I'm not sure if ProtocolExtension is the right category here (considering its disabled by everyone that wants to allow a mod on their server)

GameFlow comes close, but also not great. New category? PacketStuffing? Exploits (do we see a legit use for this?)

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
```diff
@@ -132,6 +132,18 @@ public override async ValueTask<bool> HandleRpcAsync(ClientPlayer sender, Client
                         return false;
                     }
 
+                    var position = reader.Position;
+                    var statesLength = reader.ReadPackedInt32();
+                    reader.Seek(position);
+
+                    if (statesLength < 0 || statesLength > Game.Options.MaxPlayers)
```
miniduikboot, line 139:
```suggestion
                    if (statesLength < 0 || statesLength > Game.PlayerCount)
```

Do you see a reason to allow more players than that are actually in the game?

miniduikboot (changes requested):
Using the current category, this won't take effect on most servers

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
NikoCat233: Because Innersloth told me that they use 'MaxPlayers' to fix the race condition in 'GetMaxMessagePackingLimit()'.

Since Innersloth uses 'MaxPlayers' to avoid race conditions, we could use it as well. It's a byte with a maximum value of 255, which is more than enough to cover any potential oversizing and won't cause race conditions when a player disconnects.

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
NikoCat233: Maybe a new category called Exploits, since neither normal mods nor the vanilla game would ever perform this kind of action.

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
miniduikboot: Sure, that makes sense

src/Impostor.Server/Net/Inner/Objects/InnerMeetingHud.cs
miniduikboot: Exploits is fine by me

NikoCat233:
Fixed merge conflicts

PR #735, opened 2026-06-27
Bump Hazel to 1.0.1

miniduikboot:
### Description

This pulls in https://github.com/Impostor/Impostor.Hazel/pull/4


miniduikboot:
build seems to fail because a cache somewhere hasn't picked up 1.0.1 already, will retry tomorrow

PR #738, opened 2026-08-13
Deserialize gamecode as packed in PackedGameDataTo 

miniduikboot:
### Description

Most packets use a normal 32 bit int but not PackedGameDataTo. Instead,
Innersloth decided to waste a byte by using WritePacked. As v2 game
codes are always negative, this is less efficient, but alas.

It was documented in the pseudocode and I didn't spot the difference
when reviewing the implementation, so I made this mistake when
deduplicating this code.

Test plan:
- Start EHR on at least 2 clients
- Use the game mode "Natural Disasters"
- Observe that pixelated blobs are visible for all players, not just the host

Bonus feature: on some room codes this code was crashing as the truncated 4 bytes weren't valid. The IsPacketAllowed doesn't need to actually parse the game code, it just needs to check that the value matches, so I removed that parse check

### Closes issues

- amends #726

### DIFF
--- a/src/Impostor.Server/Net/Client.cs
+++ b/src/Impostor.Server/Net/Client.cs
@@ -313,7 +313,7 @@ namespace Impostor.Server.Net
                             return;
                         }
 
-                        if (packed.ReadInt32() != Player!.Game.Code)
+                        if (packed.ReadInt32() != game.Code.Value)
                         {
                             _logger.LogWarning("PackedGameDataTo contained GameDataTo for the wrong game.");
                             return;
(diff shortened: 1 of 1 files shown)


src/Impostor.Server/Net/Client.cs
HayashiUme: Should this be changed to game.Code.Value, this an oversight?

HayashiUme (changes requested):

src/Impostor.Server/Net/Client.cs
miniduikboot: Good catch. It doesn't need to be retrieved from Player again now that we have an

Unfortunate that C# boxes the int32 instead of unboxing the Code attribute, but oh well. I'm not going to audit the entire codebase for this issue, but if it is spotted, please fix it

PR #739, opened 2026-08-18
Among Us 18.0

miniduikboot:
### Description

This adds support for Among Us 18.0 and the Judge role

Open issues:
- [x] Finish adapting MeetingHud to JudgeOverrides

- [x] Add support for [AMCI](https://github.com/Innersloth-LLC/AmongUsModdingInformation#all-client-mod-client-identification)
- [x] Fix Dumpostor as it's not returning proper enums at the moment

### Closes issues

- closes #

PR #740, opened 2026-08-19
Addition v18.0

HayashiUme:
### Description

Add a part of support for AmongUs v18.0, should merge into #739 with review

### Closes issues

- closes #

### DIFF
--- a/src/Impostor.Api/Net/Messages/C2S/Message00HostGameC2S.cs
+++ b/src/Impostor.Api/Net/Messages/C2S/Message00HostGameC2S.cs
@@ -21,7 +21,7 @@ namespace Impostor.Api.Net.Messages.C2S
             crossplayFlags = (CrossplayFlags)reader.ReadInt32();
             gameFilterOptions = GameFilterOptions.Deserialize(reader);
         }
-        
+
         public static void DeserializeModded(IMessageReader reader, out IGameOptions gameOptions, out CrossplayFlags crossplayFlags, out GameFilterOptions gameFilterOptions, out Guid modGuid)
         {
             Deserialize(reader, out gameOptions, out crossplayFlags, out gameFilterOptions);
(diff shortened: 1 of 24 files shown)


src/Impostor.Api/Net/Messages/C2S/Message00HostGameC2S.cs
```diff
@@ -20,5 +21,11 @@ public static void Deserialize(IMessageReader reader, out IGameOptions gameOptio
             crossplayFlags = (CrossplayFlags)reader.ReadInt32();
             gameFilterOptions = GameFilterOptions.Deserialize(reader);
         }
+
+        public static void DeserializeModded(IMessageReader reader, out IGameOptions gameOptions, out CrossplayFlags crossplayFlags, out GameFilterOptions gameFilterOptions, out Guid modGuid)
```
miniduikboot, line 25: You're deserializing a different message here, so it belongs in a different class

src/Impostor.Api/Net/Messages/Rpcs/Rpc23VotingComplete.cs
```diff
@@ -2,14 +2,16 @@ namespace Impostor.Api.Net.Messages.Rpcs
...
         {
             writer.WriteBytesAndSize(states);
             writer.Write(playerId);
             writer.Write(tie);
+            writer.Write(wasOverruled);
+            writer.Write(overrideId);
         }
 
-        public static void Deserialize(IMessageReader reader, out IMessageReader[] states, out byte playerId, out bool tie)
+        public static void Deserialize(IMessageReader reader, bool hasOverruleFields, out IMessageReader[] states, out byte playerId, out bool tie, out bool wasOverruled, out ushort overrideId)
```
miniduikboot, line 14: I don't think it's a good idea to pass a bool that affects parsing like this: once we don't support any pre-18.0 version anymore, we want to remove this boolean, but that's difficult as it's in the API.

I'm going to change this to an overload: one version will deserialize the 17.x version, another the 18.x version, idem for Serialize

miniduikboot (approved):
Thanks, will fix the comments in my own branch

